PR #36, opened 2019-11-28
Improve appmap testing

apotterri:
These changes rework testing somewhat, adding support for testing both ruby 2.5 and ruby 2.6. Additionally, I added a test fixture containing a Rails 4 app with the same functionality as the current Rails 5 app.

There are alot of files in this PR, mostly because of the new app. It's going to be easiest to review the commits individually. You should pay particular attention to acb2275, dbb03df, and 5979243. You can safely ignore 387e312, as it's only the result of running `rails new`, and has no other changes. 8caee45 updates the Rails 4 app so its functionality matches the Rails 5, shouldn't need more than a quick look.


Rakefile
```diff
@@ -1,16 +1,198 @@
...
+      run_specs(ruby_version, task_args)
+    end.tap do|t|
+      desc "Run all specs"
+      # multitask all: t
+      task all: t
+    end
+  end
+end
+
+=begin
```
kgilpin: Is this block meant to be deleted?

Rakefile
```diff
@@ -1,16 +1,198 @@
...
+
+namespace :spec do
+  RUBY_VERSIONS.each do |ruby_version|
+    desc ruby_version
+    task ruby_version, [:specs] => ["build:fixtures:#{ruby_version}:all"] do |_, task_args|
+      run_specs(ruby_version, task_args)
+    end.tap do|t|
+      desc "Run all specs"
+      # multitask all: t
+      task all: t
```
kgilpin: Is this going to redefine `spec:all`? I'm not sure what happens if a task is redefined with different dependencies.

spec/fixtures/rails_users_app/Dockerfile
```diff
@@ -1,6 +1,11 @@
-FROM appmap-ruby_with_appmap:2.6
+ARG GEM_VERSION
+ARG RUBY_VERSION
 
-RUN apt-get update && apt-get install -y postgresql-client
+FROM appmap:${GEM_VERSION} as appmap
+
+FROM ruby:${RUBY_VERSION}
```
kgilpin: Cool, I just learned about multi-stage builds.

Rakefile
```diff
@@ -1,16 +1,198 @@
...
+            build_app_image(app, ruby_version)
+          end.tap do |t|
+            desc "Build all fixture images for #{ruby_version}"
+            task all: t
+          end
+        end
+      end
+      
+      desc "Build all fixture images"
+      # multitask all: ["#{ruby_version}:all"]
```
kgilpin: Looks like this can be removed.

Rakefile
```diff
@@ -1,16 +1,198 @@
...
+  
+  namespace :fixtures do
+    RUBY_VERSIONS.each do |ruby_version|
+      namespace ruby_version do
+        desc "build:fixtures:#{ruby_version}"
+        FIXTURE_APPS.each do |app|
+          desc app
+          task app => ["base:#{ruby_version}"] do
+            build_app_image(app, ruby_version)
+          end.tap do |t|
```
kgilpin: Can you explain how this works, to define a task and then define another task (in this case `all), which may be defined multiple times?

Rakefile
apotterri: Redefining a task adds to its dependencies. I stumbled on this behavior while I was experimenting with how I wanted this to look. But, it turns out it's expected: https://ruby.github.io/rake/doc/rakefile_rdoc.html#label-Multiple+Definitions

`[specs]` indicates that the `spec:all` task takes an argument. In this case, you can give it a list of files that it should run. So, for example,

```
$ bundle exec rake spec:all[spec/record_sql_rails4_pg_spec.rb] 
```

will run only the new Rails 4 specs using both Ruby 2.5 and 2.6 (after rebuilding all the dependencies).

Rakefile
apotterri: Yes, done.

spec/fixtures/rails_users_app/Dockerfile
apotterri: Yeah, they can be a big help.

Rakefile
apotterri: Done.

Rakefile
apotterri: See my comment above, but let me know if that doesn't clear things up.

PR #37, opened 2019-11-28
Fix various appmap issues

apotterri:
Fixes for various appmap issues. See the comments on the commits for details.

### DIFF
--- a/lib/appmap/rspec.rb
+++ b/lib/appmap/rspec.rb
@@ -160,6 +160,26 @@ module AppMap
 
       LOG = false
 
+      def is_example_group_subclass_call?(tp)
+        # Order is important here. Checking for method_id == :subclass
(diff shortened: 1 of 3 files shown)


lib/appmap/rspec.rb
```diff
@@ -5,6 +5,47 @@
 require 'appmap/inspect'
 require 'appmap/trace/tracer'
 
+if ActiveSupport::VERSION::MAJOR < 5
+  require 'active_support/inflector/transliterate'
+  require 'active_support/core_ext/string'
+  
+  module InflectorPatch
+    def parameterize(string, separator: "-", preserve_case: false, locale: nil)
```
kgilpin: How does this differ from the existing behavior of `parameterize`? It is safe to monkey-patch it?

PR #44, opened 2019-12-11
Scenario metadata "recorder" and "client" fields

kgilpin:
Scenario metadata has `recorder` and `client` fields, to indicate how the data was collected.

Also a couple of bug fixes:

* Handle errors which occur while calling `tp.self.method(tp.method_id).parameters`
* Collect all RSpec `label` tags from the example and its parent example groups
* Improve the test coverage for inference of feature group and feature names in when recording RSpecs.

### DIFF
--- a/lib/appmap/trace/tracer.rb
+++ b/lib/appmap/trace/tracer.rb
@@ -135,7 +135,7 @@ module AppMap
         end
 
         def collect_parameters(tp)
-          tp.self.method(tp.method_id).parameters.collect do |pinfo|
+          -> { tp.self.method(tp.method_id).parameters rescue [] }.call.collect do |pinfo|
             kind, key = pinfo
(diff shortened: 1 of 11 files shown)


lib/appmap/trace/tracer.rb
```diff
@@ -135,7 +135,7 @@ def collect_self(tp)
         end
 
         def collect_parameters(tp)
-          tp.self.method(tp.method_id).parameters.collect do |pinfo|
+          -> { tp.self.method(tp.method_id).parameters rescue [] }.call.collect do |pinfo|
```
apotterri: A couple of questions about this:

* Which of the methods here might raise exceptions, and what would they be? Does it make sense to say that if an exception does get raised, that this implies that the method has no parameters?

* Why add this as a lambda, rather than a new method?

PR #46, opened 2019-12-31
Run tests on Travis

apotterri:
These changes add a build config for Travis, and make builds a little less noisy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,38 @@
+language: ruby
+
+services:
+- docker
+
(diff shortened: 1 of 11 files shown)


apotterri:
I noticed that the "checks" for this build were red for a previous build. This was because the Google Cloud Build failed. The check for the Travis build was green. So, if we remove the Cloud Build webhook, we'll get the correct build status for PRs.

kgilpin:
OK, I've removed Google Cloud Build.

PR #58, opened 2020-05-07
Fix frameworks metadata

apotterri:
Ensure that `rspec` only shows up once in `metadata[:frameworks]`.

Also, make sure that the fixture's Gemfile references the local repository for testing.


test/rspec_test.rb
```diff
@@ -9,6 +9,7 @@ def perform_test(test_name)
     Bundler.with_clean_env do
       Dir.chdir 'test/fixtures/rspec_recorder' do
         FileUtils.rm_rf 'tmp'
+        system 'bundle config --local local.appmap ../../..'
```
kgilpin, line 12: I like how you're using `bundle config` here.

kgilpin:
Nice job. One request: Update version.rb and CHANGELOG.

PR #62, opened 2020-07-06
Fix display singleton classes, only record "call" events

apotterri:
Make sure `defined_class` for a method owned by a singleton class looks like a class name, and not like Ruby's internal representation. 

Also, for `call` events, set the `static` attribute based on the method's receiver. `return` events don't know the receiver, so the `static` attribute is no longer set for them.

Source code use to generate examples:
https://gitlab.com/apotterri-groups/app.land/singleton-methods . `master` branch of that repo has a Gemfile that specifies v0.28.0 of `appmap-ruby`. `obj-singleton_20200706` branch points to this code. Mapsets were generated with `bundle exec bin/appmap record test.rb` on the appropriate branch.

Mapset of code before changes:
https://app.land/applications/111?mapset=244
profile-before

Scenario of code before changes:
https://app.land/scenarios/edc038eb-d4b1-4766-a3cb-815dc2573672#filterShow=sql_select%2Csql_insert%2Csql_update%2Cstatic_asset_requests%2Cstack_depth_less&filterShowParams=%7B%22stack_depth_less%22%3A%228%22%7D
scenario-before

Mapset of code after changes:
https://app.land/applications/111?mapset=245
profile-after

Scenario of code after changes:
https://app.land/scenarios/ce1d0c6d-53c1-4835-8804-e657a20f597d#filterShow=sql_select%2Csql_insert%2Csql_update%2Cstatic_asset_requests%2Cstack_depth_less&filterShowParams=%7B%22stack_depth_less%22%3A%228%22%7D
scenario-after


lib/appmap/algorithm/stats.rb
```diff
@@ -74,10 +74,11 @@ def perform(limit: nil)
 
         class_name_func = ->(event) { event['defined_class'] }
         full_name_func = lambda do |event|
+          call = event['event'] == 'call'
           class_name = event['defined_class']
           static = event['static']
           function_name = event['method_id']
-          [ class_name, static ? '.' : '#', function_name ].join if class_name && !static.nil? && function_name
+          [ class_name, static ? '.' : '#', function_name ].join if call && class_name && !static.nil? && function_name
```
kgilpin, line 81: Good catch.

lib/appmap/event.rb
```diff
@@ -88,17 +84,21 @@ def build_from_invocation(mc = MethodCall.new, defined_class, method, receiver,
               object_id: receiver.__id__,
               value: display_string(receiver)
             }
+            mc.static = receiver.is_a?(Module)
```
kgilpin, line 87: So this is kind of the key thing here.

PR #68, opened 2020-08-04
Specify minimal data for call and return events

kgilpin:
Remove unnecessary or redundant data in call and return events, such as `defined_class` and `method_id` on `return` events, and `path` and `lineno` on http and sql calls.


lib/appmap/trace.rb
```diff
@@ -14,34 +14,34 @@ def initialize(defined_class, method, static)
 
     class Tracing
       def initialize
-        @Tracing = []
+        @tracing = []
```
kgilpin, line 17: I don't know how this got in here...

lib/appmap/util.rb
```diff
@@ -35,6 +35,30 @@ def scenario_filename(name, max_length: 255, separator: '_', extension: '.appmap
...
+        delete_object_id.call(event[:return_value])
+        (event[:parameters] || []).each(&delete_object_id)
+        (event[:exceptions] || []).each(&delete_object_id)
+
+        case event[:event]
+        when :call
+          event[:path] = event[:path].gsub(Gem.dir + '/', '')
+        when :return
+          # These should be removed from the appmap spec
+          %i[defined_class method_id path lineno static].each do |obsolete_field|
```
apotterri, line 55: Do these still need to be here now?

kgilpin: Good catch!

PR #72, opened 2020-08-31
Don't interfere with methods that use Time.now

apotterri:
Ensure that capturing events doesn't change the behavior of a hooked
method that uses `Time.now`. For example, if a test expects that
`Time.now` will be called a certain number of times by a hooked method,
that expectation will now be met.

### DIFF
--- a/lib/appmap/hook/method.rb
+++ b/lib/appmap/hook/method.rb
@@ -6,6 +6,11 @@ module AppMap
       HOOK_DISABLE_KEY = 'AppMap::Hook.disable'
       private_constant :HOOK_DISABLE_KEY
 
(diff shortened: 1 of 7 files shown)


lib/appmap/hook/method.rb
```diff
@@ -52,12 +57,12 @@ def before_hook(receiver, args)
         require 'appmap/event'
         call_event = AppMap::Event::MethodCall.build_from_invocation(defined_class, hook_method, receiver, args)
         AppMap.tracing.record_event call_event, defined_class: defined_class, method: hook_method
-        [ call_event, Time.now ]
```
kgilpin, line 55: Heh, good catch.

PR #73, opened 2020-09-03
Add extension to fix Hook.qualify_method_name

apotterri:
Add a native extension that allows `qualify_method_name` to avoid any
functions the code being recorded may have redefined (e.g. `#to_s` or
`.name).

### DIFF
--- a/lib/appmap.rb
+++ b/lib/appmap.rb
@@ -16,6 +16,9 @@ require 'appmap/metadata'
 require 'appmap/util'
 require 'appmap/open'
 
+# load extension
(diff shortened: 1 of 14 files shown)


lib/appmap.rb
```diff
@@ -16,6 +16,9 @@
 require 'appmap/util'
 require 'appmap/open'
 
+# load extension
+require 'appmap/appmap'
```
kgilpin, line 20: It looks like this line is failing in the minitests.

kgilpin (approved):

PR #76, opened 2020-09-21
Fix problem getting database version

apotterri:
`ActiveRecord::Base.connection.server_version` was added in Rails 6.
Don't try to call it unless it's available.


kgilpin (approved):

kgilpin:
Gem is released.

PR #78, opened 2020-09-24
Get rid of extra warnings

apotterri:
Only warn once about problems determining database version for an
ActiveRecord connection.

### DIFF
--- a/lib/appmap/rails/sql_handler.rb
+++ b/lib/appmap/rails/sql_handler.rb
@@ -72,9 +72,17 @@ module AppMap
         end
 
         class ActiveRecordExaminer
+          @@db_version_warning_issued = {}
+          
+          def issue_warning
+            db_type = database_type
+            return if @@db_version_warning_issued[db_type]
+            warn("AppMap: Unable to determine database version for #{db_type.inspect}") 
+            @@db_version_warning_issued[db_type] = true
+          end
+          
           def server_version
-            ActiveRecord::Base.connection.try(:database_version) ||\
-              warn("Unable to determine database version for #{database_type.inspect}")
+            ActiveRecord::Base.connection.try(:database_version) || issue_warning
           end
 
           def database_type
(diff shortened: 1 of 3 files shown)


lib/appmap/rails/sql_handler.rb
```diff
@@ -71,21 +71,32 @@ def execute_query(sql)
...
+              attr_reader :server_version, :database_type
+              
+              @database_type = begin
+                type = ActiveRecord::Base.connection.adapter_name.downcase.to_sym
+                type = :postgres if type == :postgresql
+                
+                type
+              end
+              
+              @server_version = ActiveRecord::Base.connection.try(:database_version) or
```
kgilpin, line 86: Will this work if the app is using multiple different connections?

lib/appmap/rails/sql_handler.rb
```diff
@@ -71,21 +71,32 @@ def execute_query(sql)
...
-        class ActiveRecordExaminer
-          def server_version
-            ActiveRecord::Base.connection.try(:database_version) ||\
-              warn("Unable to determine database version for #{database_type.inspect}")
-          end
-
-          def database_type
-            type = ActiveRecord::Base.connection.adapter_name.downcase.to_sym
-            type = :postgres if type == :postgresql
+        if defined?(ActiveRecord)
```
kgilpin, line 74: We are only using this class if ActiveRecord is defined:

https://github.com/applandinc/appmap-ruby/blob/42917c73c7a2459caee420072544376bd26c76b2/lib/appmap/rails/sql_handler.rb#L54

So this line seems redundant with that? Could anything go wrong if this file is loaded before ActiveRecord is required?

lib/appmap/rails/sql_handler.rb
apotterri: Nope. I'll figure out a different approach that handles that case, too.

lib/appmap/rails/sql_handler.rb
apotterri: Yeah, this is no good, either.

kgilpin (approved):

PR #79, opened 2020-09-24
Enhance hooks with mime type and custom display string

kgilpin:
Provide custom display string for certain types.

Report `mime_type` on HTTP server response.


lib/appmap/event.rb
```diff
@@ -36,20 +36,38 @@ def display_string(value)
...
+        def custom_display_string(value)
+          case value
+          when File
+            "#{value.class}[path=#{value.path}]"
+          when Net::HTTP
+            "#{value.class}[#{value.address}:#{value.port}]"
+          when Net::HTTPGenericRequest
+            "#{value.class}[#{value.method} #{value.path}]"
+          end
+        rescue StandardError
```
apotterri, line 55: `StandardError` is what gets caught by a bare `rescue`, isn't it? Is there some reason it needs to be mentioned explicitly here?

apotterri (approved):
One question, looks good on the whole.

lib/appmap/event.rb
kgilpin: Rubucop :-(

PR #82, opened 2020-09-30
Specify a gem name to record in appmap.yml

kgilpin:
Enable appmap.yml to specify a `gem` name to map the gem.

Example:

```
name: gem_test
packages:
  - gem: activesupport
```

### DIFF
--- a/lib/appmap/config.rb
+++ b/lib/appmap/config.rb
@@ -4,16 +4,25 @@ module AppMap
   class Config
     Package = Struct.new(:path, :gem, :package_name, :exclude, :labels) do
       class << self
-        def build(path: nil, gem: nil, package_name: nil, exclude: [], labels: [])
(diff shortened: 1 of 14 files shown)


lib/appmap/config.rb
```diff
@@ -2,15 +2,30 @@
 
 module AppMap
   class Config
-    Package = Struct.new(:path, :package_name, :exclude, :labels) do
-      def initialize(path, package_name: nil, exclude: [], labels: [])
-        super path, package_name, exclude, labels
+    Package = Struct.new(:path, :gem, :package_name, :exclude, :labels) do
```
kgilpin, line 5: A Package might now be constructed from a `gem`. If so, record that fact here.

apotterri (approved):
LGTM

PR #83, opened 2020-10-01
Fix `MethodEvent#display_string`

apotterri:
Make sure `MethodEvent#display_string` can handle values with
problematic `#to_s` and `#inspect` methods.


dustinbyrne (approved):
👍

PR #84, opened 2020-10-01
Allow skipping railtie if APPMAP_INITIALIZE=false

dividedmind:
Also, don't track `.ruby-version`; it's a gem and should be developed and tested with different Ruby versions.


kgilpin (changes requested):
In order to release this, it would be best to update the version and changelog

dividedmind:
> In order to release this, it would be best to update the version and changelog

I updated the changelog but I didn't think the change is worth a release and a version bump all in itself, especially since nothing depends on it. I figured it'd be included in a future bump.

kgilpin:
Ok, as long as you aren’t relying on anyone picking up / not picking up
this change.

kgilpin (approved):

PR #85, opened 2020-10-02
fix params mapping

flywithmemsl:
This PR fixes next issue with params mapping:
image

image

I'm not sure if it's nice solution, but it solves the problem and I would be glad to hear comments how can it be improved.

### DIFF
--- a/lib/appmap/event.rb
+++ b/lib/appmap/event.rb
@@ -89,15 +89,17 @@ module AppMap
             else
               mc.path = [ defined_class, static ? '.' : '#', method.name ].join
             end
+            parameters_length = method.parameters.length
+            arguments_length = arguments.length
             mc.parameters = method.parameters.map.with_index do |method_param, idx|
               param_type, param_name = method_param
               param_name ||= 'arg'
-
-              if (arguments.length == 1) && (arguments[0].is_a? Hash) && (arguments[0][param_name])
-                value = arguments[0][param_name]
-              else
-                value = arguments[idx]
-              end
+              value =
+                if (arguments[0].is_a? Hash) && (arguments[0][param_name]) && (parameters_length != 1) && (arguments_length == 1)
+                  value = arguments[0][param_name]
+                else
+                  value = arguments[idx]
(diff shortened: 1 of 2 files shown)


lib/appmap/event.rb
```diff
@@ -92,7 +92,12 @@ def build_from_invocation(mc = MethodCall.new, defined_class, method, receiver,
             mc.parameters = method.parameters.map.with_index do |method_param, idx|
               param_type, param_name = method_param
               param_name ||= 'arg'
-              value = arguments[idx]
+
+              if (arguments.length == 1) && (arguments[0].is_a? Hash) && (arguments[0][param_name])
+                value = arguments[0][param_name]
+              else
+                value = arguments[idx]
+              end
```
dividedmind, line 100: Hash isn't always the first argument, but it's always the last one. You should look at the parameter type instead, and use `arguments[-1]` if the type is `:key` or `:keyreq`. (Not sure what'll be in `arguments` if all keyword arguments are optional and none is provided; hopefully an empty hash in the last place. If not, you'll have to pop the posititonal arguments to determine if keyword arguments are provided.)

dividedmind (changes requested):
This will fail in most cases; see specific comment near the code. Please start with adding and modifying specs to test what you expect to obtain; then you can work on the code to accomplish it (or you could ask someone else). Be sure to test all the variations (default, keyword and positional arguments in all the different permutations).

lib/appmap/event.rb
flywithmemsl: It wasn't supposed to be solving a hash argument problem. It was a fix only for single edge case when params are collected it a wrong way. So I check it and fix on the fly. But thanks to your comment I examined it one more time and improved `if` condition to make it work more accurate.

dividedmind:
Hi @flywithmemsl , thanks for the change, I've added a bit more comprehensive implementation at https://github.com/applandinc/appmap-ruby/pull/89 -- could you take a look and double check if it solves the problem? Thanks!

PR #86, opened 2020-10-08
Capture method source code and comment

dividedmind:
This change builds on https://github.com/applandinc/appmap/pull/9 to capture method source code and documentation comments in the classmap. Note for RSpec runs to avoid duplication code is only recorded in the inventory file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ appmap.json
 .vscode
 .byebug_history
 /lib/appmap/appmap.bundle
+*.so
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,6 @@
+AllCops:
+  NewCops: enable
+
 Layout/SpaceInsideArrayLiteralBrackets:
   Enabled: false
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v0.37.0
+* Capture method source and comment.
+
 # v0.36.0
 * *appmap.yml* package definition may specify `gem`.
 * Skip loading the railtie if `APPMAP_INITIALIZE` environment variable
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,4 @@
+$: << File.join(__dir__, 'lib')
 require 'appmap/version'
 GEM_VERSION = AppMap::VERSION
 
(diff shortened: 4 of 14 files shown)


dividedmind:
Sorry about the failing tests. I'm fixing them, but they seem only a matter of changing the expected JSON to include the source, so feel free to review the code in the meantime.

apotterri:
These changes will reintroduce the requirement to upload the `Inventory.json` file, right? I believe the server doesn't currently require this, it just uses the classmaps from the appmaps -- the land-of-apps builds don't upload the inventory file, and their diagrams are fully populated.

Additionally, `Inventory.json` gets overwritten each time you run the client. So, if you run a subset of the specs, then run a second subset, the source will only get saved for the second subset. You'll have appmaps for all the specs, though.

I understand why you saved the source in the inventory file rather than in the appmaps, but there might be more that needs to be done.

kgilpin:
It might be better to just drop the Inventory optimization (because that's
all it is), for these reasons. Even though it will mean a lot of
duplication in the appmap files. It's like Russian rockets; big, heavy, but
robust.

dividedmind:
I really don't have a good answer for this. I thought this is good enough but if we find it confusing it's simple to switch it off. I took care to make the server code robust enough to handle whatever you throw at it. So tell me: should I add the inventory optimization to minitest, or remove it from rspec? Up to you. My vote is to keep it (and add it to minitest) for now, FWIW.

(BTW, I'm surprised that land-of-apps builds don't upload it. I kinda assumed they just uploaded *.json.)

apotterri:
Yeah, I misspoke about the land-of-apps builds. It does upload `*.json`, because that's what `appland upload` does. This is less than ideal, though, because the server doesn't handle the Inventory file specially. Instead, it winds up looking like a scenario without any events in it.

So, it seems like we should drop the optimization, unless we can show that it's really necessary, and we fix the management of the Inventory file.

kgilpin:
Inventory.json is weird, but that is not the fault of this PR :-) So after thinking about it, I think we should merge this as-is.

dividedmind:
> This is less than ideal, though, because the server doesn't handle the Inventory file specially. Instead, it winds up looking like a scenario without any events in it.

It does handle it specially: it hides it in the scenario list. That's just about enough.

kgilpin (approved):

PR #94, opened 2020-12-29
Record "view" rendering, and implement API "surface" recording

kgilpin:
* Implement API "surface" recording. When `shallow: true`, only the first function call entering a package is recorded. Subsequent functions within the package are not recorded. When recording gems using the `gem: <gem_name>` option in appmap.yml, shallow recording is the default.
* Record `ActionView::Renderer#render` automatically, and apply the label `view`.
* Update the gem version to `1.0.0`.

Regarding the tests on Travis ; `appmap-ruby` is broken on Travis for some reason. I have sent a note to their support. Meanwhile, I ran the test suite locally on a clean checkout and it passed.

### DIFF
--- a/lib/appmap/trace.rb
+++ b/lib/appmap/trace.rb
@@ -34,7 +34,7 @@ module AppMap
       end
 
       def last_package_for_current_thread
-        @tracers.map(&:last_package_for_current_thread).first
+        @tracers.first&.last_package_for_current_thread
       end
 
       def record_event(event, package: nil, defined_class: nil, method: nil)
--- a/lib/appmap/trace.rb
+++ b/lib/appmap/trace.rb
@@ -15,34 +15,38 @@ module AppMap
 
     class Tracing
       def initialize
-        @tracing = []
+        @tracers = []
       end
 
       def empty?
-        @tracing.empty?
+        @tracers.empty?
       end
 
       def trace(enable: true)
         Tracer.new.tap do |tracer|
-          @tracing << tracer
+          @tracers << tracer
           tracer.enable if enable
         end
       end
 
       def enabled?
-        @tracing.any?(&:enabled?)
+        @tracers.any?(&:enabled?)
+      end
+
+      def last_package_for_current_thread
+        @tracers.map(&:last_package_for_current_thread).first
       end
 
       def record_event(event, package: nil, defined_class: nil, method: nil)
-        @tracing.each do |tracer|
+        @tracers.each do |tracer|
           tracer.record_event(event, package: package, defined_class: defined_class, method: method)
         end
       end
 
       def delete(tracer)
-        return unless @tracing.member?(tracer)
+        return unless @tracers.member?(tracer)
 
-        @tracing.delete(tracer)
+        @tracers.delete(tracer)
         tracer.disable
       end
     end
@@ -52,6 +56,7 @@ module AppMap
     # Records the events which happen in a program.
     def initialize
       @events = []
+      @last_package_for_thread = {}
       @methods = Set.new
       @enabled = false
     end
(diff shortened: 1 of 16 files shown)


lib/appmap/trace.rb
```diff
@@ -15,34 +15,38 @@ def initialize(package, defined_class, method, static)
...
         end
       end
 
       def enabled?
-        @tracing.any?(&:enabled?)
+        @tracers.any?(&:enabled?)
+      end
+
+      def last_package_for_current_thread
+        @tracers.map(&:last_package_for_current_thread).first
```
dividedmind, line 37: Is it the same as
```suggestion
        @tracers.first&.last_package_for_current_thread
```
but slower? ;)

spec/abstract_controller_base_spec.rb
```diff
@@ -74,9 +75,28 @@
...
+  method_id: render
+  path: .*
+  lineno: .*
+  static: false
+          VIEW_CALL
+
+          expect(appmap).to match(<<-VIEW_LABEL.strip)
+          "labels":["view"]
+          VIEW_LABEL
+        end
```
dividedmind: Can squiggly heredocs be used here so it's indented nicely?

spec/fixtures/rails6_users_app/Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '~> 5.2.3'
+gem 'rails', '~> 6'
```
dividedmind, line 5: Nice catch ;)

spec/abstract_controller_base_spec.rb
```diff
@@ -3,61 +3,65 @@
...
-          appmap = JSON.parse(File.read(create_user_appmap_json)).to_yaml
+          it 'properly captures method parameters in the appmap' do
+            expect(File).to exist(appmap_json_file)
+            appmap = JSON.parse(File.read(appmap_json_file)).to_yaml
 
-          expect(appmap).to match(<<-CREATE_CALL.strip)
+            expect(appmap).to match(<<-CREATE_CALL.strip)
   event: call
   thread_id: .*
   defined_class: Api::UsersController
```
dividedmind: Might as well make those heredocs squiggly while you're at it.

spec/abstract_controller_base_spec.rb
```diff
@@ -72,33 +76,38 @@
...
+            expect(event.keys).to eq(%w[id event thread_id parent_id elapsed return_value])
+          end
         end
-
-        it 'records and labels view rendering' do
-          expect(File).to exist(list_users_appmap_json)
-          appmap = JSON.parse(File.read(list_users_appmap_json)).to_yaml
-
-          expect(appmap).to match(<<-VIEW_CALL.strip)
+        describe 'listing users' do
```
dividedmind, line 89: An empty line should go between the describe blocks.

spec/hook_spec.rb
```diff
@@ -741,9 +741,33 @@ def say_default
...
+        :class: String
+        :value: string
+        :kind: :req
+      - :name: :b
+        :class: String
+        :value: string
+        :kind: :req
+      :receiver:
+        :class: Module
+        :value: ActiveSupport::SecurityUtils
```
dividedmind, line 769: I'd indent this one step more.

dividedmind (approved):
Looks good, thanks! I've left a couple of style suggestions, take them or leave them. Please just be sure to reorder commits before merging -- some of them should become fixups and the (bold!) version bump should come last.

spec/abstract_controller_base_spec.rb
kgilpin: The indentation matters when matching YAML as a string. Probably a nicer way to do it, but ...

dividedmind: But these are unindented at the base level, no? And squiggly heredocs remove the extra indentation.

PR #98, opened 2021-01-21
Add Ruby 2.7 to the test matrix

kgilpin:
Add support for Ruby 2.7, and remove support for Rails 4 (which is deprecated).


.travis.yml
```diff
@@ -16,22 +16,8 @@ before_script:
...
-    script:
-    - bundle exec rake build:base:2.6
-      
-  - stage: spec
-    script:
-    - bundle exec rake spec:2.5
-  - stage: spec
-    script:
-    - bundle exec rake spec:2.6
+    - bundle exec rake test
```
apotterri, line 22: This is certainly simpler. :)

I wonder if there's value in running (at least) the minitest tests in multiple Ruby versions. [This](https://docs.travis-ci.com/user/build-stages/#build-stages-and-build-matrix-expansion) suggests that using matrix expansion for a single stage is pretty straightforward. I'll leave it up to you.

apotterri (approved):
Once the build is green again, this is good to go.

PR #104, opened 2021-01-29
Labels from comments

kgilpin:
# Labels

The [AppMap data format](https://github.com/applandinc/appmap) provides for class and function `labels`, which can be used to enhance the AppMap visualizations, and to programatically analyze the data.

You can apply function labels using source code comments in your Ruby code. To apply a labels to a function, add a `@label` or `@labels` line to the comment which immediately precedes a function.

For example, if you add this comment to your source code:

```ruby
class ApiKey
  # @labels provider.authentication security
  def authenticate(key)
    # logic to verify the key here...
  end
end
```

Then the AppMap metadata section for this function will include:

```json
  {
    "name": "authenticate",
    "type": "function",
    "labels": [ "provider.authentication", "security" ]
  }
```

### DIFF
--- a/lib/appmap/class_map.rb
+++ b/lib/appmap/class_map.rb
@@ -113,17 +113,18 @@ module AppMap
             [ method.defined_class, static ? '.' : '#', method.name ].join
           end
 
-        if include_source
-          begin
-            function_info[:source] = method.source
-            comment = method.comment || ''
-            function_info[:comment] = comment unless comment.empty?
-          rescue MethodSource::SourceNotFoundError
-            # pass
-          end
+        source, comment = begin
+          [ method.source, method.comment ]
+        rescue MethodSource::SourceNotFoundError
+          [ nil, nil, ]
         end
 
-        function_info[:labels] = package.labels if package.labels
+        if include_source
+          function_info[:source] = source unless source.blank?
+          function_info[:comment] = comment unless comment.blank?
+        end
+
+        function_info[:labels] = parse_labels(comment) + (package.labels || [])
         object_infos << function_info
 
         parent = root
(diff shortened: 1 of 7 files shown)


lib/appmap/class_map.rb
```diff
@@ -141,6 +142,22 @@ def add_function(root, method, include_source: true)
...
+      # Labels can be embedded in the function comment. Label format is similar to YARD and JavaDoc.
+      # The keyword is @labels or @label. The keyword is followed by space-separated labels.
+      # For example:
+      # @label provider.authentication security
+      def parse_labels(comment)
+        return [] unless comment
+
+        comment
+          .split("\n")
+          .map { |line| line.match(/^\s*#\s*@labels?\s+(.*)/) }
```
apotterri, line 154: Is there any value to using something like https://ruby.github.io/rdoc/RDoc/Text.html#method-i-normalize_comment to ensure that the comment is going to be formatted in a way we'll understand?

spec/spec_helper.rb
```diff
@@ -3,6 +3,7 @@
 require 'json'
 require 'yaml'
 require 'English'
+require 'byebug'
```
apotterri, line 6: Was this addition intentional?

apotterri (approved):
Love this idea. Changes look good.

lib/appmap/class_map.rb
kgilpin: Hmm, I guess there may be some comment forms that I don’t know about? I’m just covering the common case 

```ruby
  # @labels foo bar
```

spec/spec_helper.rb
kgilpin: Yes I like it being available so I can add `byebug` anywhere to get a debugger. Is it a bad practice?

lib/appmap/class_map.rb
apotterri: Probably fine for now. We may want to revisit if we find that there are other forms we need to handle -- it might make sense to leverage some other work, rather than making this code more complicated.

spec/spec_helper.rb
apotterri: I have a key bound to insert `require 'pry'; binding.pry`. :)

A linter would probably complain about an unused `require`, but given that it's in `spec/spec_helper.rb`, it doesn't seem like it matters very much.

kgilpin:
Ok cool. I’m gonna merge them :-)

PR #111, opened 2021-03-30
Enhance HTTP server request recording

kgilpin:
Adds more useful information to `http_server_request` and `_response`. This enhances our ability to generate Swagger.

* Record `name` and `class` of each entry in Hash-like parameters, messages, and return values.

Example: 
```ruby
                      'name' => 'user',
                      'properties' => [
                        { 'name' => 'login', 'class' => 'String' },
                        { 'name' => 'password', 'class' => 'String' }
                      ]
```

* Record client-sent headers in HTTP server request and response.

Example:
```ruby
                  'http_server_request' => hash_including(
                    'headers' => hash_including('X-Sandwich' => 'turkey')
                  )
```

* Record HTTP server request `mime_type` (client `Content-Type` header).
* Record HTTP server request `authorization` (client `Authorization` header).

Fix specs by including only `sequel` or `active_record` database cleaner.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # v0.43.0
 
-* **Experimental** Record `name` and `class` of each entry in Hash-like parameters, messages, and return values.
-* **Experimental** Record most headers in HTTP server request and response.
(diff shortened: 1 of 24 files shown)


kgilpin:
Specs are fixed!!!

PR #114, opened 2021-04-25
Refine config

### LINKED ISSUE #115: When running with Minitest, raises wrong number of arguments (given 2, expected 3)
It looks like [this change](https://github.com/applandinc/appmap-ruby/pull/109) broke things for Minitest. The `save` method [takes 3 position args](https://github.com/applandinc/appmap-ruby/blob/master/lib/appmap/minitest.rb#L81), but [here it is called with only two](https://github.com/applandinc/appmap-ruby/blob/10eb2a5ce49634a3a479e0d9c1737b050911e473/lib/appmap/minitest.rb#L109).

kgilpin:
* Support recording and labeling of indivudal functions via `functions:` section in *appmap.yml*.
* Remove deprecated `exe/appmap`.
* Add `test_status` and `exception` fields to AppMap metadata.
* Write AppMap file atomically, by writing to a temp file first and then moving it into place.
* Remove printing of `Inventory.json` file.
* Remove source code from `classMap`.

Fixes #115


README.md
```diff
@@ -145,6 +155,11 @@ Each entry in the `packages` list is a YAML object which has the following keys:
 
 Optional list of fully qualified class and method names. Separate class and method names with period (`.`) for class methods and hash (`#`) for instance methods.
 
+**functions**
+
+Optional list of `class, function` pairs. The `package` name is used to place the function within the class map, and does not have to match
+the folder or gem name. The primary use of `functions` is to apply specific labels to functions whose source code is not accessible (e.g., it's in a Gem).
+For functions which are part of the application code, using `@label` or `@labels` in code comments to apply labels.
```
apotterri, line 162: This should be "**use** `@label` or `@labels`", yes?

lib/appmap/hook/method.rb
```diff
@@ -38,6 +38,14 @@ def activate
           warn "AppMap: Hooking #{msg} at line #{(hook_method.source_location || []).join(':')}"
         end
 
+        call_method_name = :call
```
apotterri, line 41: There should probably be a test for this special case.

apotterri (changes requested):
A couple of minor changes to make. Could use some history cleanup, too.

apotterri (approved):

PR #117, opened 2021-04-29
Record HTTP client requests and responses

kgilpin:
Implement HTTP client request/response, per https://github.com/applandinc/appmap/pull/18 (and adding `message` to the request).


.travis.yml
```diff
@@ -16,19 +16,6 @@ before_script:
...
-# https://stackoverflow.com/a/41975912
-before_cache:
-  # Save tagged docker images
-  - >
-    mkdir -p $HOME/docker && docker images -a --filter='dangling=false' --format '{{.Repository}}:{{.Tag}} {{.ID}}'
-    | xargs -n 2 -t sh -c 'test -e $HOME/docker/$1.tar.gz || docker save $0 | gzip -2 > $HOME/docker/$1.tar.gz'
-
-before_install:
-  # Load cached docker images
-  - if [[ -d $HOME/docker ]]; then ls $HOME/docker/*.tar.gz | xargs -I {file} sh -c "zcat {file} | docker load"; fi
```
kgilpin, line 31: This was only slowing the build down even more.

lib/appmap/handler/net_http.rb
```diff
@@ -0,0 +1,107 @@
...
+            h[:message] = params.keys.map do |key|
+              val = params[key]
+              {
+                name: key,
+                class: val.class.name,
+                value: self.class.display_string(val),
+                object_id: val.__id__,
+              }
+            end
+          end
```
dividedmind, line 51: This won't have complete URL then, because the query parameters end up in `message`, right? Is it more convenient this way? Should the spec proposal be revised?

BTW, it doesn't make any sense to put `object_id` or `class` there, they're all going to be strings that have just been created. I know the spec requires it, but perhaps we should make them optional.

kgilpin:
Yes, it's more convenient with "message", because the tools and UI already
know how to handle "message". They don't know how to parse a URL. We could
strip the query string from the URL. We don't care about anchor tags anyway.

lib/appmap/handler/net_http.rb
```diff
@@ -0,0 +1,107 @@
...
+
+require 'appmap/event'
+
+module AppMap
+  module Handler
+    class HTTPClientRequest < AppMap::Event::MethodEvent
+      attr_accessor :request_method, :url, :params, :headers
+
+      def initialize(http, request)
+        super AppMap::Event.next_id_counter, :call, Thread.current.object_id
```
dividedmind, line 11: It looks like the superclass initialization wants to take care of the ids.

lib/appmap/handler/net_http.rb
```diff
@@ -0,0 +1,107 @@
...
+        protocol = http.use_ssl? ? 'https' : 'http'
+        port = if http.use_ssl? && http.port == 443
+          nil
+        elsif !http.use_ssl? && http.port == 80
+          nil
+        else
+          ":#{http.port}"
+        end
+
+        url = [ protocol, '://', http.address, port, request.path ].compact.join
```
dividedmind, line 25: Let's remove the query if it's going to be passed as params. I've updated the https://github.com/applandinc/appmap/pull/18 proposal.

lib/appmap/rails/request_handler.rb
```diff
@@ -59,16 +39,18 @@ def to_h
...
-              {
-                name: key,
-                class: val.class.name,
-                value: self.class.display_string(val),
-                object_id: val.__id__,
-              }.tap do |message|
-                properties = object_properties(val)
-                message[:properties] = properties if properties
+            unless params.blank?
+              h[:message] = params.keys.map do |key|
```
dividedmind, line 43: I've always wondered why it's `message`. Why not just pass HTTP parameters in the `parameters` attribute?

lib/appmap/util.rb
```diff
@@ -72,6 +80,19 @@ def sanitize_event(event, &block)
...
+        matching_headers = env
+          .select { |k,v| k.start_with? 'HTTP_'}
+          .reject { |k,v| v.blank? }
+          .each_with_object({}) do |kv, memo|
+            key = kv[0].sub(/^HTTP_/, '').split('_').map(&:capitalize).join('-')
+            value = kv[1]
+            memo[key] = value
+          end
+        matching_headers.blank? ? nil : matching_headers
+      end
```
dividedmind, line 94: Doesn't this just reimplement `request.headers` now (with the exception of niling if empty)?

spec/record_net_http_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+  end
+end
+
+describe 'Net::HTTP handler' do
+  include_context 'collect events'
+
+  def get_hello
+    http = Net::HTTP.new('localhost', 19292)
+    http.get '/hello?msg=hi'
```
dividedmind, line 25: Can we test with multiple query parameters, some duplicated and some urlencoded?

dividedmind (approved):
Looks good, I've just had some comments.

lib/appmap/rails/request_handler.rb
kgilpin: `message` is understood by the UI code.

PR #122, opened 2021-05-10
ci-performance: support optional Dockerhub auth again

hleb-rubanau:
(bring back the previous change after it was unintentionally deleted)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,6 @@ before_script:
 
 cache:
   bundler: true
-  directories:
-    - $HOME/docker
 
 before_install:
   # see https://blog.travis-ci.com/docker-rate-limits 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -16,6 +16,18 @@ before_script:
 
 cache:
   bundler: true
+  directories:
```
apotterri, line 19: Do we need to be caching `$HOME/docker`? I don't see anything getting put there.

apotterri (changes requested):

.travis.yml
hleb-rubanau: Certainly not. It propagated from another changeset which should be removed. Will fix in a moment.

apotterri (approved):

PR #123, opened 2021-05-13
feat: Emit swagger-style normalized paths instead of Rails-style ones

kgilpin:


lib/appmap/handler/rails/request_handler.rb
```diff
@@ -65,7 +65,7 @@ def normalized_path(request, router = ::Rails.application.routes.router)
               next unless app.matches? request
               return normalized_path request, app.rack_app.routes.router if app.engine?
 
-              return route.path.spec.to_s
+              return AppMap::Util.swaggerize_path(route.path.spec.to_s)
```
apotterri, line 68: FWIW, I looked at this a bit, too. As an alternative to parsing the result of `to_s`, you could use `each`: https://github.com/rails/rails/blob/v6.1.3.2/actionpack/lib/action_dispatch/journey/nodes/node.rb#L18 .

apotterri (approved):

lib/appmap/handler/rails/request_handler.rb
kgilpin: OK. Do you want to push an update or should we merge as-is?

apotterri: I can push an update. Should I also fix the build failures?

kgilpin: Sure, if you can :-) Not sure why those are going on.

apotterri: Using a Visitor turned out to be more trouble than it was worth. I fixed the tests, I'll merge it when the build's done.

PR #125, opened 2021-05-17
Conditional init

kgilpin:
Initialization (code hooking) only proceeds if `APPMAP=true`.

Omitting `APPMAP=true` enables quick and clean startup, and normal debugging, even in test and development.


apotterri (approved):
LGTM

PR #134, opened 2021-06-16
feat: Add refinement to the labels

kgilpin:
Differentiates between some `read` and `write` operations.


apotterri (approved):
We should maybe add some tests for these, at some point.

kgilpin:
> We should maybe add some tests for these, at some point.

Yes, there are some tests for some of these labels, but testing every single one seems ... well it's more than I've been willing to do so far :-)

PR #142, opened 2021-06-18
feat: Swagger and depends tasks

kgilpin:
Incorporates rake tasks to generate Swagger and update AppMaps based on file modification times.

Bundles the NPM package `@appland/cli`.


spec/fixtures/rails5_users_app/docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
     build:
       context: .
       dockerfile: Dockerfile
-    image: rails-app:${RUBY_VERSION}
+    image: rails5-app:${RUBY_VERSION}
```
kgilpin, line 16: @apotterri this actually fixes a bug in which the same Docker image (`rails-app`) was being used for both Rails5 and Rails6 testing. Which ever image was built last would be use by both test suites. So with this change, now we are reliably testing both Rails 5 and 6.

spec/fixtures/rails6_users_app/docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
     build:
       context: .
       dockerfile: Dockerfile
-    image: rails-app:${RUBY_VERSION}
+    image: rails6-app:${RUBY_VERSION}
```
kgilpin, line 16: This is the other half of the Docker fixture image fix for Rails 5/6.

kgilpin:
Split out into separate PRs.

PR #143, opened 2021-06-18
feat: Provide default appmap.yml settings

kgilpin:
If appmap.yml is not found, warn the user.
Detect some possible source paths, and the project name from the .git file,
and use this info to create a temporary appmap.yml.
Print it to the console so the user can use it as a starting point.

Screen Shot 2021-06-18 at 3 05 22 PM

### DIFF
--- a/lib/appmap.rb
+++ b/lib/appmap.rb
@@ -27,7 +27,7 @@ module AppMap
     # Gets the configuration. If there is no configuration, the default
     # configuration is initialized.
     def configuration
-      @configuration ||= initialize
+      @configuration ||= initialize_configuration
     end
 
     # Sets the configuration. This is only expected to happen once per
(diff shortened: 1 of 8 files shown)


lib/appmap.rb
```diff
@@ -38,12 +38,15 @@ def configuration=(config)
       @configuration = config
     end
 
+    def default_config_file_path
+      ENV['APPMAP_CONFIG_FILE'] || 'appmap.yml'
+    end
+
     # Configures AppMap for recording. Default behavior is to configure from "appmap.yml".
```
apotterri, line 45: This comment could probably use an update.

lib/appmap/config.rb
```diff
@@ -253,38 +257,118 @@ def initialize(name, packages, exclude: [], functions: [])
...
+        config_params = {
+          exclude: config_data['exclude']
+        }.compact
+
+        if config_data['functions']
+          config_params[:functions] = config_data['functions'].map do |function_data|
+            package = function_data['package']
+            cls = function_data['class']
+            functions = function_data['function'] || function_data['functions']
+            raise 'AppMap class configuration should specify package, class and function(s)' unless package && cls && functions
```
apotterri, line 321: Should this say "AppMap **function** configuration should specify..."?

apotterri (approved):

PR #145, opened 2021-06-21
fix: Fix problems with auto-requiring appmap modules

kgilpin:
`if defined?(::Rails::Railtie)` doesn't reliably return true, due to load order issues. So, we also use TracePoint to watch modules being loaded.

This PR also removes redundant methods from the ClassMap, by ensuring that method objects are well-behaved in a Set.

### DIFF
--- a/lib/appmap.rb
+++ b/lib/appmap.rb
@@ -112,19 +112,57 @@ module AppMap
   end
 end
 
-if Gem.loaded_specs['rails']
-  require 'active_support'
-  require 'active_support/core_ext'
-  require 'rails'
-  require 'appmap/railtie' 
-end
+lambda do
+  Initializer = Struct.new(:class_name, :module_name, :gem_module_name)
 
-if Gem.loaded_specs['rspec-core']
-  require 'appmap/rspec'
-end
+  INITIALIZERS = {
+    'Rails::Railtie' => Initializer.new('AppMap::Railtie', 'appmap/railtie', 'railtie'),
+    'RSpec' => Initializer.new('AppMap::RSpec', 'appmap/rspec', 'rspec-core'),
+    'Minitest::Unit::TestCase' => Initializer.new('AppMap::Minitest', 'appmap/minitest', 'minitest')
+  }
+
+  print_msg = if defined?(::Rails) && Rails.logger
+    Rails.logger
+  elsif ENV['DEBUG'] == 'true'
+    Object.method(:warn)
+  else
+    ->(msg) {}
+  end
+
+  TracePoint.new(:class) do |tp|
+    cls_name = tp.self.name
+    initializers = INITIALIZERS.delete(cls_name)
+    if initializers
+      initializers = [ initializers ] unless initializers.is_a?(Array)
+      next if Object.const_defined?(initializers.first.class_name)
+
+      gem_module_name = initializers.first.gem_module_name
+
+      print_msg.call AppMap::Util.color(<<~LOAD_MSG, :magenta)
+      When 'appmap' was loaded, '#{gem_module_name}' had not been loaded yet. Now '#{gem_module_name}' has
+      just been loaded, so the following AppMap modules will be automatically required:
+
+      #{initializers.map(&:module_name).join("\n")}
+
+      To suppress this message, ensure '#{gem_module_name}' appears before 'appmap' in your Gemfile.
+      LOAD_MSG
+      initializers.each do |init|
+        require init.module_name
+      end
+    end
+  end.enable
+
+  if defined?(::Rails::Railtie)
+    require 'appmap/railtie'
+  end
   
-if Gem.loaded_specs['minitest']
-  require 'appmap/minitest'
-end
+  if defined?(::RSpec)
+    require 'appmap/rspec'
+  end
   
-AppMap.initialize_configuration if ENV['APPMAP'] == 'true'
+  if defined?(::Minitest)
+    require 'appmap/minitest'
+  end
+  
+  AppMap.initialize_configuration
+end.call if ENV['APPMAP'] == 'true'
--- a/lib/appmap.rb
+++ b/lib/appmap.rb
@@ -51,7 +51,7 @@ module AppMap
     # Call this function before the program code is loaded by the Ruby VM, otherwise
     # the load events won't be seen and the hooks won't activate.
     def initialize_configuration(config_file_path = default_config_file_path)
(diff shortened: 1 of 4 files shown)


lib/appmap.rb
```diff
@@ -112,17 +112,19 @@ def detect_metadata
   end
 end
 
-if defined?(::Rails::Railtie)
```
kgilpin, line 115: Due to load order, `::Rails::Railtie` may well not be defined here; even if Rails is in the bundle.

lib/appmap.rb
```diff
@@ -112,17 +112,19 @@ def detect_metadata
...
+if Gem.loaded_specs['rails']
+  require 'active_support'
+  require 'active_support/core_ext'
+  require 'rails'
   require 'appmap/railtie' 
 end
 
-if defined?(::RSpec)
+if Gem.loaded_specs['rspec-core']
```
kgilpin, line 122: Probably best to use the same approach here.

lib/appmap/handler/rails/template.rb
```diff
@@ -14,16 +14,30 @@ class Template
...
+            [ package, path, name ]
+          end
+
+          def hash
+            id.hash
+          end
+
+          def eql?(other)
+            id.eql? other.id
+          end
```
kgilpin, line 39: These Method objects are placed into a Set in order to make them unique. To behave properly in a set, `hash` and `eql?` are required.

lib/appmap/trace.rb
```diff
@@ -2,10 +2,12 @@
 
 module AppMap
   module Trace
-    class RubyMethod
+    class RubyMethod < SimpleDelegator
```
kgilpin, line 5: To behave properly in a Set, inherit `hash` and `eql?` from the `method` parameter.

lib/appmap.rb
```diff
@@ -112,17 +112,49 @@ def detect_metadata
...
+      #{initializers.map(&:module_name).join("\n")}
+
+      To suppress this message, require '#{gem_module_name}' before you require 'appmap' in your Gemfile.
+      LOAD_MSG
+      initializers.each do |init|
+        require init.module_name
+      end
+    end
+  end.enable
+end.call
```
dividedmind, line 146: Why the immediately called lambda?

lib/appmap.rb
```diff
@@ -112,17 +112,49 @@ def detect_metadata
...
+    'Rails::Railtie' => Initializer.new('AppMap::Railtie', 'appmap/railtie', 'railtie'),
+    'RSpec' => Initializer.new('AppMap::RSpec', 'appmap/rspec', 'rspec-core'),
+    'Minitest::Unit::TestCase' => Initializer.new('AppMap::RSpec', 'appmap/minitest', 'minitest')
+  }
+
+  TracePoint.new(:class) do |tp|
+    cls_name = tp.self.name
+    initializers = INITIALIZERS.delete(cls_name)
+    if initializers
+      initializers = [ initializers ] unless initializers.is_a?(Array)
```
dividedmind, line 128:
```suggestion
      initializers = Array(initializers)
```

Also, none of them is an array currently, so maybe there's no need to handle multiple?

lib/appmap.rb
```diff
@@ -112,17 +112,49 @@ def detect_metadata
   end
 end
 
+lambda do
+  Initializer = Struct.new(:class_name, :module_name, :gem_module_name)
+
+  INITIALIZERS = {
+    'Rails::Railtie' => Initializer.new('AppMap::Railtie', 'appmap/railtie', 'railtie'),
+    'RSpec' => Initializer.new('AppMap::RSpec', 'appmap/rspec', 'rspec-core'),
+    'Minitest::Unit::TestCase' => Initializer.new('AppMap::RSpec', 'appmap/minitest', 'minitest')
```
dividedmind, line 121: Is `AppMap::RSpec` here correct?

lib/appmap.rb
```diff
@@ -112,17 +112,49 @@ def detect_metadata
...
   require 'appmap/rspec'
 end
 
-# defined?(::Minitest) returns nil...
-if Gem.loaded_specs['minitest']
+if defined?(::Minitest)
   require 'appmap/minitest'
 end
-
+
```
dividedmind, line 159: You remove trailing spaces in one place then add them in another... balance, eh? ;)

dividedmind (approved):
- Thinking more about this, I wonder if we should show the message at all. It made sense if there was a way to prevent autoloading (the message would note that autoload wasn't performed), but since we're already doing it anyway maybe it's not very useful after all.
- The `require if defined?` clauses below can be driven off INITIALIZERS so it's DRYer, eg. `INITIALIZERS.filter(&Object.method(:const_defined?)).values.map(:module_name).each(&Kernel.method(:require)`.
- If ENV['APPMAP'] is not true, maybe we ought to skip the whole thing?

lib/appmap.rb
kgilpin: No, thanks.

lib/appmap.rb
kgilpin: In another branch, there is a multiple.

lib/appmap.rb
kgilpin: To prevent INITIALIZERS from being in the global namespace. Am I right that could happen? I could refactor to avoid this.

lib/appmap.rb
kgilpin: Ugh. I need some tool to take care of this for me.

lib/appmap.rb
```diff
@@ -112,19 +112,57 @@ def detect_metadata
...
+      gem_module_name = initializers.first.gem_module_name
+
+      print_msg.call AppMap::Util.color(<<~LOAD_MSG, :magenta)
+      When 'appmap' was loaded, '#{gem_module_name}' had not been loaded yet. Now '#{gem_module_name}' has
+      just been loaded, so the following AppMap modules will be automatically required:
+
+      #{initializers.map(&:module_name).join("\n")}
+
+      To suppress this message, require '#{gem_module_name}' before you require 'appmap' in your Gemfile.
```
apotterri, line 150: I think it's worth rewording this message to avoid the use of the word `require` -- because `require` is  Ruby keyword, and isn't actually used in a Gemfile, I thing this message may lead to confusion. You could say something like
```
ensure '#{gem_module_name}' appears before 'appmap' in your Gemfile.
```

apotterri (changes requested):

PR #146, opened 2021-06-22
feat: Add CLI setup agent tool with `init` command 

torchello:


exe/appmap
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env ruby
```
kgilpin, line 1: I think we should call this program something more descriptive, perhaps `appmap-agent-setup`, just to be very clear what it's for, and to avoid users calling it by accident by just typing `appmap` for whatever reason.

exe/appmap
```diff
@@ -0,0 +1,49 @@
...
+# frozen_string_literal: true
+
+require 'gli'
+
+require 'appmap/version'
+require 'appmap/hook'
```
kgilpin, line 7: I see `AppMap::Config` being used here, but not `AppMap::Hook`.

exe/appmap
```diff
@@ -0,0 +1,49 @@
...
+
+class App
+  extend GLI::App
+
+  program_desc 'AppMap CLI'
```
kgilpin, line 12: Maybe more descriptive about the full intent of this program.

exe/appmap
```diff
@@ -0,0 +1,49 @@
...
+  command :init do |c|
+    c.action do
+      require 'appmap/command/init'
+      AppMap::Command::Init.new(@config_file).perform
+    end
+  end
+
+  pre do |global, _, _, _|
+    @config_file = global[:config]
+    @config = interpret_config_option(@config_file) if File.exist?(@config_file)
```
kgilpin, line 35: Is `@config` used for anything?

test/cli_test.rb
```diff
@@ -0,0 +1,38 @@
...
+  CONFIG_FILENAME = '123.yml'
+  SUBFOLDER_CONFIG_FILEPATH = 'conf/123.yml'
+  EXPECTED_CONFIG_CONTENT = %(name: appmap-ruby
+packages:
+- path: lib
+)
+
+  def test_init_when_config_exists
+    output = `./exe/appmap init`
+    p output
```
kgilpin, line 16: Stray line here :-)

exe/appmap
torchello: Not now. It will be used by other subcommands.

exe/appmap
torchello: Required `appmap` instead both.

kgilpin (approved):

lib/appmap/service/guesser.rb
```diff
@@ -0,0 +1,26 @@
...
+    class Guesser
+      POSSIBLE_PATHS = %w[app/controllers app/models lib]
+      class << self
+        def guess_name
+          reponame = lambda do
+            next unless File.directory?('.git')
+
+            repo_name = `git config --get remote.origin.url`.strip
+            repo_name.split('/').last.split('.').first unless repo_name == ''
+          end
```
apotterri, line 14: There are a couple of problems with this way of guessing the repo name: the default remote won't necessarily be  named `origin`, and the repo name might have a `.` in it (e.g. https://github.com/chicagopython/chipy.org).

I think would be better to just use the name of the root directory of the repo, i.e. `basename $(git rev-parse --show-toplevel)`.

PR #149, opened 2021-06-23
feat: Add swagger rake task

kgilpin:
Client usage looks like this:

```
namespace :appmap do
  AppMap::Swagger::RakeTasks.define_tasks
end
```

Configuration options are available in *appmap.yml*.

Also incorporates work to reduce dependency on activesupport, by providing simple implementations of functions such as `blank?`, `classify`, and `deep_dup`.


lib/appmap/agent.rb
```diff
@@ -0,0 +1,115 @@
...
+require_relative 'trace'
+require_relative 'class_map'
+require_relative 'metadata'
+require_relative 'util'
+require_relative 'open'
+
+# load extension
+require_relative 'appmap'
+
+module AppMap
```
apotterri, line 13: Was this just extracted verbatim from lib/appmap.rb? Are there any changes in here that need to be reviewed>

apotterri (approved):
One question, but LGTM.

lib/appmap/agent.rb
kgilpin: The only changes was to use `Util.deep_dup` instead of ActiveSupport core_ext `deep_dup`. In Util, it does a 2-way Marshal instead of traversing through all the data. Since that works fine for this case and performance doesn't much matter.

PR #150, opened 2021-06-23
fix: Better project name guesser

torchello:


lib/appmap/service/guesser.rb
```diff
@@ -6,15 +6,8 @@ class Guesser
...
-          reponame = lambda do
-            next unless File.directory?('.git')
-
-            repo_name = `git config --get remote.origin.url`.strip
-            repo_name.split('/').last.split('.').first unless repo_name == ''
-          end
-          dirname = -> { Dir.pwd.split('/').last }
-
-          reponame.() || dirname.()
+          return `basename $(git rev-parse --show-toplevel)`.strip if File.directory?('.git')
```
kgilpin, line 9: We should use a Ruby function (`Pathname`?) rather than `basename` because basename may not be available on all platforms (e.g. Windows).

apotterri: Yeah. My suggestion was meant to show what we wanted to do, not how it should be implemented.

torchello: Good catch! Fixed.

kgilpin (approved):

Issue #153, opened 2021-06-25
ArgumentError proxying kwrest arguments under certain conditions


kgilpin:
In Ruby 2.7, the following call sequence will produce an ArgumentError:

```ruby
class Kwargs
  class << self
    def no_kwargs(args)
      args
    end

    def has_kwrest_calls_no_kwargs(args, **kwargs)
      no_kwargs(**kwargs)
    end
  end
end
```

```
     Failure/Error:
       def no_kwargs(args)
         args
       end
     
     ArgumentError:
       wrong number of arguments (given 0, expected 1)
     # spec/fixtures/hook/kwargs.rb:3:in `no_kwargs'
     # ./lib/appmap/hook/method.rb:77:in `block (2 levels) in activate'
     # ./lib/appmap/hook/method.rb:100:in `block in activate'
     # spec/fixtures/hook/kwargs.rb:20:in `has_kwrest_calls_no_kwargs'
```

PR #154, opened 2021-06-25
fix: Workaround Ruby bug in 2.7.3 with kwrest

kgilpin:
There are two main elements to this fix:

* Call https://ruby-doc.org/core-2.7.0/Proc.html#method-i-ruby2_keywords
* Fix a probable side-effect of using ruby_keywords: https://github.com/applandinc/appmap-ruby/pull/154/files#diff-95a15f6e0d63580f2d31bf5a2e866f1062fbda51da73ad9089c579b952467a08R54

Fixes https://github.com/applandinc/appmap-ruby/issues/153

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: ruby
 
+rbenv:
+- 2.5
+- 2.6
+- 2.7
+
 addons:
   apt:
     packages:
@@ -38,7 +43,6 @@ jobs:
     - mkdir tmp
     - GEM_ALTERNATIVE_NAME='' bundle exec rake test
 
-    
 before_deploy:
   - |
     nvm install --lts \
--- a/Rakefile
+++ b/Rakefile
@@ -6,8 +6,7 @@ require 'rake/testtask'
 require 'rdoc/task'
 
 require 'open3'
-
-require "rake/extensiontask"
+require 'rake/extensiontask'
 
 desc 'build the native extension'
 Rake::ExtensionTask.new("appmap") do |ext|
@@ -26,7 +25,7 @@ namespace 'gem' do
       # ~/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/gem_helper.rb:39:in `install'
       def build_gem
         # Ensure that NPM packages are installed before building.
-        sh('yarn install --prod'.shellsplit)
+        sh('yarn install --prod')
   
         default_build_gem
       end
@@ -34,7 +33,17 @@ namespace 'gem' do
   end
 end
 
-RUBY_VERSIONS=%w[2.5 2.6 2.7]
+RUBY_VERSIONS=%w[2.5 2.6 2.7].select do |version|
+  travis_ruby_version = ENV['TRAVIS_RUBY_VERSION']
+  next true unless travis_ruby_version
+
+  if travis_ruby_version.index(version) == 0
+    warn "Testing Ruby version #{version}, since it matches TRAVIS_RUBY_VERSION=#{travis_ruby_version}"
+    next true
+  end
+
+  false
+end
 FIXTURE_APPS=%w[rack_users_app rails6_users_app rails5_users_app]
 
 def run_cmd(*cmd)
--- a/appmap.gemspec
+++ b/appmap.gemspec
@@ -54,4 +54,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'webdrivers', '~> 4.0'
   spec.add_development_dependency 'timecop'
   spec.add_development_dependency 'hashie'
+  spec.add_development_dependency 'webrick'
 end
--- a/lib/appmap/handler/net_http.rb
+++ b/lib/appmap/handler/net_http.rb
@@ -82,7 +82,7 @@ module AppMap
         def request_headers(request)
           {}.tap do |headers|
             request.each_header do |k,v|
-              key = [ 'HTTP', k.underscore.upcase ].join('_')
+              key = [ 'HTTP', Util.underscore(k).upcase ].join('_')
               headers[key] = v
             end
           end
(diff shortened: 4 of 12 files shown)


.travis.yml
```diff
@@ -1,5 +1,10 @@
 language: ruby
 
+rbenv:
```
kgilpin, line 3: We go to a build matrix scheme in order to run the non-Dockerized tests across Ruby versions.

Rakefile
```diff
@@ -26,15 +25,25 @@ namespace 'gem' do
...
+        sh('yarn install --prod')
   
         default_build_gem
       end
     end
   end
 end
 
-RUBY_VERSIONS=%w[2.5 2.6 2.7]
+RUBY_VERSIONS=%w[2.5 2.6 2.7].select do |version|
```
kgilpin, line 36: This is the simplest way to get the behavior we want without doing a major build refactor.

appmap.gemspec
```diff
@@ -54,4 +54,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'webdrivers', '~> 4.0'
   spec.add_development_dependency 'timecop'
   spec.add_development_dependency 'hashie'
+  spec.add_development_dependency 'webrick'
```
kgilpin, line 57: This is a separate Gem in Ruby 3. Preparing for that.

lib/appmap/handler/net_http.rb
```diff
@@ -82,7 +82,7 @@ class << self
         def request_headers(request)
           {}.tap do |headers|
             request.each_header do |k,v|
-              key = [ 'HTTP', k.underscore.upcase ].join('_')
+              key = [ 'HTTP', Util.underscore(k).upcase ].join('_')
```
kgilpin, line 85: Moving away from activesupport.

lib/appmap/handler/rails/template.rb
```diff
@@ -146,7 +146,8 @@ def handle_return(call_event_id, elapsed, return_value, exception)
         class RenderHandler
           class << self
             def handle_call(defined_class, hook_method, receiver, args)
-              context, options = args
+              # context, options
```
kgilpin, line 149: Fix an unused argument warning.

lib/appmap/hook.rb
```diff
@@ -99,7 +99,6 @@ def trace_location(trace_point)
     end
 
     def trace_end(trace_point)
-      location = trace_location(trace_point)
```
kgilpin, line 102: Fix an unused argument warning.

lib/appmap/hook/method.rb
```diff
@@ -45,42 +47,48 @@ def activate
...
-        hook_method_def = nil
-        hook_class.instance_eval do 
-          hook_method_def = Proc.new do |*args, &block|
-            instance_method = hook_method.bind(self).to_proc
-            call_instance_method = -> { instance_method.call(*args, &block) }
+        hook_method_def = Proc.new do |*args, &block|
+          instance_method = hook_method.bind(self).to_proc
+          call_instance_method = -> {
+            # https://github.com/applandinc/appmap-ruby/issues/153
+            if Util.ruby_minor_version >= 2.7 && args == ARRAY_OF_EMPTY_HASH && hook_method.arity == 1
```
kgilpin, line 54: This is the actual fix for the bug. In this particular case, `instance_method.call(*args, &block)` will actually pass 0 args to the instance_method. We detect that case and make the correct call.

lib/appmap/hook/method.rb
```diff
@@ -45,42 +47,48 @@ def activate
...
+          rescue
+            exception = $ERROR_INFO
+            raise
+          ensure
+            with_disabled_hook.call do
+              after_hook.call(self, call_event, start_time, return_value, exception) if call_event
             end
           end
         end
+        hook_method_def = hook_method_def.ruby2_keywords if Util.ruby_minor_version >= 2.7
```
kgilpin, line 90: This is probably how the problem above is introduced - but it's necessary to have the proper behavior in Ruby 2.7.

lib/appmap/trace.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'delegate'
```
kgilpin, line 3: Turns out this is supposed to be a separate import.

lib/appmap/util.rb
```diff
@@ -124,7 +124,7 @@ def swaggerize_path(path)
         path = path.split('(.')[0]
         tokens = path.split('/')
         tokens.map do |token|
-          token.gsub /^:(.*)/, '{\1}'
+          token.gsub(/^:(.*)/, '{\1}')
```
kgilpin, line 127: This was a warning.

spec/record_net_http_spec.rb
```diff
@@ -1,6 +1,7 @@
 require 'spec_helper'
 require 'diffy'
 require 'rack'
+require 'webrick'
```
kgilpin, line 4: Webrick is a separate gem in Ruby 3.0.

lib/appmap/hook/method.rb
apotterri: The doc suggests using `respond_to?(:ruby2_keywords)`. Is there any reason to prefer that?

apotterri: So, you're saying this causes a problem, and the change above works around the problem, right? Can you give an example of the bad behavior this change fixes, please?

apotterri (commented):
I'd like understand the results of your investigation and how the 2.7-specific changes work a bit better. Other changes here look good.

lib/appmap/hook/method.rb
kgilpin: So, without calling `ruby2_keywords`, then in the case above, args is `[]` and not `[{}]`, so there's really no way to know what the argument to the proxied method should be. At least with `ruby2_keywords`, it's possible to determine what should be happening.

kgilpin: We can use `respond_to?`

torchello (approved):
Looks good! Suggestion: should we try keeping PRs smaller for easier review? (separate fixes in separate PRs)

kgilpin:
Large PRs are a problem, but to me this isn’t that big.

apotterri (approved):

PR #155, opened 2021-06-25
fix: Require appmap/railtie if Rails is defined

kgilpin:
require 'appmap/railtie' will trigger autoload of 'rails/railtie', which
will allow AppMap::Railtie to work properly.

require 'appmap/railtie' when the class Rails::Railtie is first observed
does not work, because other classes such as
Rails::Railtie::Configuration are not yet available.

So, check if Rails is defined. If so, load 'appmap/railtie'. In a Rails
app, Rails is always defined when the dependency gems are loaded,
because Rails startup begins by loading 'rails/commands'.

### DIFF
--- a/lib/appmap.rb
+++ b/lib/appmap.rb
@@ -26,7 +26,8 @@ lambda do
   Initializer = Struct.new(:class_name, :module_name, :gem_module_name)
 
   INITIALIZERS = {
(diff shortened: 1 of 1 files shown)


apotterri (approved):
LGTM

PR #157, opened 2021-06-25
feat: add appmap-agent-status executable with config/project properties

torchello:


exe/appmap-agent-status
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'optparse'
+require 'appmap'
+require 'appmap/command/agent_setup/status'
+
+@options = {:config_file => AppMap::DEFAULT_CONFIG_FILE_PATH}
```
kgilpin, line 8: We can use `{ config_file: value }` style.

exe/appmap-agent-status
```diff
@@ -0,0 +1,19 @@
...
+
+OptionParser.new do |parser|
+  parser.banner = 'Usage: bundle exec exe/appmap-agent-status [options]'
```
kgilpin, line 11: Let's explain the usage simply as `Usage: appmap-agent-status [options]`. Bundler is optional and the user may be running the command from a binstub or executable file installed by Rubygems.

lib/appmap/service/config_analyzer.rb
```diff
@@ -0,0 +1,34 @@
...
+      def initialize(config_file)
+        @config_file = config_file
+        @config = load_config
+      end
+
+      def app_name
+        @config.to_h[:name] if is_present?
+      end
+
+      def is_present?
```
kgilpin, line 17: You can just use `present?` instead of `is_present?`, `is` implies `?`.

lib/appmap/service/config_analyzer.rb
```diff
@@ -0,0 +1,34 @@
...
+        File.exist?(@config_file)
+      end
+
+      def is_valid?
```
kgilpin, line 21: Same here ; `valid?`.

lib/appmap/service/config_analyzer.rb
```diff
@@ -0,0 +1,34 @@
...
+        is_present? && @config.to_h.key?(:name) && @config.to_h.key?(:packages)
```
kgilpin, line 22: Looking at this, I think that if we are going to indicate that the config file is not valid, we need to provide a reason. So, this should be added to the status spec.

The reasons I can think of:
```
1) File is not valid YAML
2) Name is missing
3) No packages are listed (but this is not really an error, the user may want to do that on purpose?)
4) The YAML is valid, but an error occurred loading the config file: {message}
```

lib/appmap/service/config_analyzer.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+
+      private
+
+      def load_config
+        AppMap::Config.load_from_file @config_file if is_present?
```
kgilpin, line 28: Here we can implement the different error conditions that might occur. So, first check if the file is present. Then try and parse it as YAML. If it's valid YAML, then call `AppMap::Config.load`, and rescue any errors.

lib/appmap/command/agent_setup/status.rb
```diff
@@ -0,0 +1,39 @@
...
+            :properties => {
+              :config => {
+                :app => config_analyzer.app_name,
+                :present => config_analyzer.is_present?,
+                :valid => config_analyzer.is_valid?
+              },
+              :project => {
+                :agentVersionProject => AppMap::VERSION,
+                :language => 'ruby',
+                :remoteRecordingCapable => defined?(::Rails) ? true : false,
```
kgilpin, line 22: I don't think that `defined?(::Rails)` will be reliable, because Rails might be in the bundle, but not loaded. Try looking for `Gem.loaded_specs['rails']` https://www.rubydoc.info/stdlib/rubygems/Gem.loaded_specs

kgilpin: Of course we need to ensure that the code editor runs this script as `bundle exec <script-name>`, so that the whole bundle is present.

exe/appmap-agent-status
kgilpin: Also, `exe/appmap-agent-status` is only correct when running the command from the gem root directory; but the user will normally be running it from their own project root directory. Since the `appmap` gem is installed, `bundle exec appmap-agent-status` should work.

lib/appmap/service/config_analyzer.rb
torchello: Right! 👍

lib/appmap/service/config_analyzer.rb
torchello: Will be implemented in `appmap-agent-validate`.

lib/appmap/service/config_analyzer.rb
```diff
@@ -0,0 +1,34 @@
...
+      def app_name
+        @config.to_h[:name] if present?
+      end
+
+      def present?
+        File.exist?(@config_file)
+      end
+
+      def valid?
+        present? && @config.to_h.key?(:name) && @config.to_h.key?(:packages)
```
kgilpin, line 22: Isn't it the case that `present?` can be true but `@config` is nil?

kgilpin (approved):

lib/appmap/service/config_analyzer.rb
torchello: Should not be a problem as `nil.to_h` gives an empty hash:
```
irb(main):001:0> nil.to_h
=> {}
```

kgilpin:
OK. I didn't know that :-)

PR #162, opened 2021-06-29
feat: Update init command to return JSON

torchello:


lib/appmap/command/agent_setup/init.rb
```diff
@@ -11,32 +10,19 @@ module AgentSetup
...
-            :green
-          )
-          puts content
-        end
-
-        private
+          result = {
+            :configuration => {
+              :filename => config_file,
+              :contents => YAML.dump(config).gsub("---\n", '')
```
apotterri, line 21: Why are you stripping out the separator(s)?

lib/appmap/command/agent_setup/init.rb
```diff
@@ -11,32 +10,19 @@ module AgentSetup
...
-          puts AppMap::Util.color(
-            %(The following AppMap config file #{config_file} has been created:),
-            :green
-          )
-          puts content
-        end
-
-        private
+          result = {
+            :configuration => {
```
apotterri, line 19: The rest of appmap-ruby uses new-style syntax for hashes with symbol keys. Probably best to follow that convention here.

apotterri (changes requested):

lib/appmap/command/agent_setup/init.rb
torchello: We won't have multiple "documents" in one yaml, so it does not make sense here. Would you prefer to keep it?

apotterri: Unless the separators actually cause a problem, I think we should keep them.

apotterri (changes requested):
Updated hashes look good, let's just get rid of the `gsub` that strips out  the separators.

lib/appmap/command/agent_setup/init.rb
torchello: Removed!

apotterri (approved):

PR #164, opened 2021-07-05
feat: Add `test_commands` sections to `appmap-agent-status` executable

torchello:
Example output for `rails6_users_app`:
```json
{
  "test_commands": [
    {
      "framework": "rspec",
      "command": "APPMAP=true bundle exec rspec spec/controllers"
    },
    {
      "framework": "minitest",
      "command": "APPMAP=true bundle exec rails test test/controllers test/integration"
    },
    {
      "framework": "cucumber",
      "command": "APPMAP=true bundle exec cucumber"
    }
  ],
  "properties": {
    "config": {
      "app": "rails6_users_app",
      "present": true,
      "valid": true
    },
    "project": {
      "agentVersion": "0.57.0",
      "language": "ruby",
      "remoteRecordingCapable": true,
      "integrationTests": true
    }
  }
}
```

### DIFF
--- a/lib/appmap/service/integration_test_path_finder.rb
+++ b/lib/appmap/service/integration_test_path_finder.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'appmap/service/test_framework_detector'
+
+module AppMap
+  module Service
+    class IntegrationTestPathFinder
+      class << self
+        def find
+          @paths ||= begin
+            paths = { rspec: [], minitest: [], cucumber: [] }
+            paths[:rspec] = find_rspec_paths if TestFrameworkDetector.rspec_present?
+            paths[:minitest] = find_minitest_paths if TestFrameworkDetector.minitest_present?
+            paths[:cucumber] = find_cucumber_paths if TestFrameworkDetector.cucumber_present?
+            paths
+          end
+        end
+
+        def count_paths
+          find.flatten(2).length - 3
+        end
+
+        private
+
+        def find_rspec_paths
+          find_non_empty_paths(%w[spec/controllers spec/requests spec/integration spec/api spec/features spec/system])
+        end
+
+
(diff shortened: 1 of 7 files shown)


lib/appmap/service/test_command_provider.rb
```diff
@@ -0,0 +1,54 @@
...
+        end
+
+        private
+
+        def minitest_command
+          if Gem.loaded_specs.has_key?('rails')
+            "APPMAP=true bundle exec rails test #{integration_test_paths[:minitest].join(' ')}"
+          else
+            subcommands = integration_test_paths[:minitest].map { |path| "bundle exec ruby #{path}" }
+            "APPMAP=true #{subcommands.join(' && ')}"
```
apotterri, line 44: This will only set `APPMAP=true` for the first command:
```shell
bash-5.1$ FOO=bar bash -c 'echo first foo \"$FOO\"' && bash -c 'echo second foo \"$FOO\"'
first foo "bar"
second foo ""
bash-5.1$ FOO=first bash -c 'echo first foo \"$FOO\"' && FOO=second bash -c 'echo second foo \"$FOO\"'
first foo "first"
second foo "second"
```

apotterri (changes requested):
Only small change to make, looks good otherwise.

apotterri (approved):

lib/appmap/service/integration_test_path_finder.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+
+        def count_paths
+          find.flatten(2).length - 3
+        end
+
+        private
+
+        def find_rspec_paths
+          find_non_empty_paths(%w[spec/controllers spec/requests spec/integration spec/api spec/features])
```
dustinbyrne, line 26: `spec/system` is another good source of integration tests that we should capture.

PR #170, opened 2021-07-07
feat: add agent-setup-validate command

torchello:


lib/appmap/service/validator/config_validator.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        def initialize(config_file)
+          @config_file = config_file
+          @violations = []
+          @config = load_config
+        end
+
+        def valid?
+          validate_config_presence
+          @violations.length == 0
```
kgilpin, line 19: Equivalently, `@violations.empty?`

lib/appmap/service/validator/config_validator.rb
```diff
@@ -0,0 +1,50 @@
...
+        end
+
+        private
+
+        def present?
+          File.exist?(@config_file)
+        end
+
+        def load_config
+          AppMap::Config.load_from_file(@config_file) if present?
```
kgilpin, line 29: It's possible for other errors besides parse errors to come from this call. So, I'd suggest to split this into two phases.

First, `YAML.parse` the config and see if it's valid. If it's valid, `load(config_data)` and handle any exceptions (which would be unexpected, but might happen).

lib/appmap/service/validator/config_validator.rb
```diff
@@ -0,0 +1,50 @@
...
+        rescue StandardError => e
+          @violations << Violation.error(
+            filename: @config_file,
+            message: 'AppMap configuration is not valid YAML',
```
kgilpin, line 33: The config file might be valid YAML here, but some other error occurred. That's why I suggest splitting this in two.

lib/appmap/service/validator/config_validator.rb
```diff
@@ -0,0 +1,50 @@
...
+        attr_reader :config, :violations
+
+        def initialize(config_file)
+          @config_file = config_file
+          @violations = []
+          @config = load_config
+        end
+
+        def valid?
+          validate_config_presence
```
kgilpin, line 18: Are you going to check the Ruby version as well?

torchello: Planned to implement in a follow up PR, but it seems to be simple, so adding now.

lib/appmap/service/validator/config_validator.rb
torchello: Makes sense, done!

kgilpin (approved):

PR #173, opened 2021-07-09
feat: Include children `controllers` and `integration` folders into minitest paths

torchello:
@ptrdvrk found an app where tests are located in `test/internal/app/controllers`. We can easily include children dirs into the list of possible paths.

### DIFF
--- a/spec/service/integration_test_path_finder_spec.rb
+++ b/spec/service/integration_test_path_finder_spec.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'appmap/service/integration_test_path_finder'
+
+describe AppMap::Service::IntegrationTestPathFinder do
+  subject { described_class }
+
+  describe '.count' do
+    it 'counts existing paths' do
+      expect(subject.count_paths).to be(0)
+    end
+  end
+
+  describe '.find' do
+    it 'finds paths' do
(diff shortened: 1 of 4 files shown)


apotterri (changes requested):
I don't see any tests for `service`? This might be a good time to add some....

torchello:
@apotterri The services are mostly covered by integration tests for init/status commands. It's tricky to cover them by unit tests, but I agree we should have at least some smoke specs. Please check again.

spec/service/integration_test_path_finder_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'appmap/service/integration_test_path_finder'
+
+describe AppMap::Service::IntegrationTestPathFinder do
+  subject { described_class }
+
+  describe '.count' do
+    it 'counts existing paths' do
```
apotterri, line 10: These tests could pass even if the code is broken, couldn't they?

Might be better to use something like https://github.com/fakefs/fakefs to verify that `service` is really working as expected.

apotterri (changes requested):

spec/service/integration_test_path_finder_spec.rb
torchello: @apotterri I adjusted the service so it can be tested with unit tests. Please have a look.

apotterri (approved):
Looks good, thanks.

PR #178, opened 2021-07-16
fix: add `DISABLE_SPRING` flag by default to test commands

torchello:


lib/appmap/service/test_command_provider.rb
```diff
@@ -61,7 +59,7 @@ def minitest_commands
                 command: {
```
kgilpin, line 59: At this point we are just “praying” that this will work because outside of Rails we can’t rely on much standardization. Hopefully the user has an opportunity to change/refine/control this command because it’s not too likely to work on its own.

torchello: Right, they will be able to edit a command to run.

apotterri (approved):
LGTM

PR #189, opened 2021-09-23
feat: Builtin code such as Ruby Logger can be hooked via appmap.yml

kgilpin:
Add a config line in the functions section such as this:

```
- name: logger/Logger#add
  builtin: true
  label: log
```

Note that the 'name' syntax  with the fully-qualified function name is
also a new addition in this commit.

### DIFF
--- a/lib/appmap/config.rb
+++ b/lib/appmap/config.rb
@@ -117,13 +117,14 @@ module AppMap
     # entry in appmap.yml. When the Config is initialized, each Function is converted into
     # a Package and TargetMethods. It's called a Function rather than a Method, because Function
     # is the AppMap terminology.
-    Function = Struct.new(:package, :cls, :labels, :function_names) do # :nodoc:
+    Function = Struct.new(:package, :cls, :labels, :function_names, :builtin) do # :nodoc:
       def to_h
         {
           package: package,
           class: cls,
           labels: labels,
-          functions: function_names.map(&:to_sym)
+          functions: function_names.map(&:to_sym),
+          builtin: builtin
         }.compact
       end
     end
@@ -244,7 +245,7 @@ module AppMap
       @depends_config = depends_config
       @hook_paths = Set.new(packages.map(&:path))
       @exclude = exclude
-      @builtin_hooks = BUILTIN_HOOKS
+      @builtin_hooks = BUILTIN_HOOKS.dup
       @functions = functions
 
       @hooked_methods = METHOD_HOOKS.each_with_object(Hash.new { |h,k| h[k] = [] }) do |cls_target_methods, hooked_methods|
@@ -254,7 +255,14 @@ module AppMap
       functions.each do |func|
         package_options = {}
         package_options[:labels] = func.labels if func.labels
-        @hooked_methods[func.cls] << TargetMethods.new(func.function_names, Package.build_from_path(func.package, **package_options))
+        package_options[:package_name] = func.package if func.builtin
+        hook = TargetMethods.new(func.function_names, Package.build_from_path(func.package, **package_options))
+        if func.builtin
+          @builtin_hooks[func.cls] ||= []
+          @builtin_hooks[func.cls] << hook
+        else
+          @hooked_methods[func.cls] << hook
+        end
       end
 
       @hooked_methods.each_value do |hooks|
@@ -300,6 +308,7 @@ module AppMap
           MISSING_FILE_MSG
           {}
         end
+
         load(config_data).tap do |config|
           config_yaml = {
             'name' => config.name,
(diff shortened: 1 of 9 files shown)


apotterri (changes requested):
Sorry to be a nag, but this could really use a test or two.

lib/appmap/config.rb
```diff
@@ -324,15 +335,22 @@ def load(config_data)
...
+              cls = function_data['class']
+              functions = function_data['function'] || function_data['functions']
+              raise %q(AppMap config 'function' element should specify 'package', 'class' and 'function' or 'functions') unless package && cls && functions
+            end
 
             functions = Array(functions).map(&:to_sym)
             labels = function_data['label'] || function_data['labels']
             labels = Array(labels).map(&:to_s) if labels
-            Function.new(package, cls, labels, functions)
+            Function.new(package, cls, labels, functions, function_data['builtin'], function_data['require'])
```
apotterri, line 353: How did you figure out you needed to add `require`? Does it deserve a test case, too?

lib/appmap/util.rb
```diff
@@ -21,6 +21,16 @@ module Util
...
 
     class << self
+      def parse_function_name(name)
+        package_tokens = name.split('/')
+
+        class_and_name = package_tokens.pop
+        class_name, function_name, static = class_and_name.include?('.') ? class_and_name.split('.', 2) + [ true ] : class_and_name.split('#', 2) + [ false ]
+
+        raise "Malformed fully-qualified function name #{name}" unless function_name
+        [ package_tokens.empty? ? 'ruby' : package_tokens.join('/'), class_name, static, function_name ]
```
dividedmind, line 31: Why `'ruby'`? Could it just be `nil`?

lib/appmap/config.rb
dividedmind: I too wonder why and when would it be needed.

lib/appmap/util.rb
dividedmind, line 31: Why `'ruby'`? It's a bit unexpected. Could it be `nil`?

spec/fixtures/rails6_users_app/appmap.yml
```diff
@@ -5,6 +5,10 @@ packages:
   - path: app/controllers
     labels: [ mvc.controller ]
   - gem: sequel
+functions:
+  - name: logger/Logger::LogDevice#write
+    builtin: true
```
dividedmind, line 10: I don't entirely understand what makes this function a built-in and why it needs special handling. It's in the standard library, but it's not a native function (at least in my version of the MRI).

dividedmind (commented):
I had some questions about it. I'm also not sure about the `name` name (it's a bit overloaded -- it's more than a name!), but I can't think of a better alternative ;)

lib/appmap/util.rb
kgilpin: I am worried that if I do that I will break something else downstream. There hasn't been a situation where package could be nil, so I don't see a compelling reason to allow it now.

spec/fixtures/rails6_users_app/appmap.yml
kgilpin: There has pretty much always been a distinction in appmap-ruby between "builtins" and code that is hooked as it's loaded.

The primary way that appmap-ruby instruments code is by hooking into the TracePoint events that fire as source files are loaded (in response to the `require` chain). But, there is some code that's already loaded before appmap-ruby is initialized. There will never be any TracePoint events for this code. So, the only way to access it (that I know of) is to `require` it and then enumerate the methods on the classes/modules.

To say the same thing a bit differently: We instruct the developer to place the `appmap` gem at the top of the Gemfile, before any other gem. This ensures that all subsequent gems will be observed by the appmap TracePoint handler. But any code that's *not* in a gem - in other words, it's built into Ruby - may already be loaded before appmap is loaded - therefore we can't rely on TracePoint events, and we need to require-then-enumerate that code.

Most of the time, the user is only interested in hooking their own code, and gems. But sometimes, it's useful to hook a builtin - like openssl or logging.

dividedmind: In other words: if the class might conceivably be already loaded the user has to use `builtin`? This is a bit confusing, because it's not what builtin means in ruby world and the line isn't always clear. Why should the user know or care about this?

Can we instead simply check for the existence of the target? If it exists, then handle it as "builtin", if not, then add a hook pending future load. (I'll also note that `builtin` path always works, even for code which is in gems.)

kgilpin: I get it, but I don't have a great alternative. Because I don't want to go `require`-ing every gem that the user lists in the appmap.yml - because it's entirely possible that a gem has `require: false`, and I am going to screw up the app if I `require` that gem.

dividedmind: Why would you need to require it? Either it's already been loaded, and then go ahead and hook is as `builtin`, otherwise add a pending hook that'll fire whenever the module is required in the normal flow. I don't understand why we'd ever need to explicitly load anything here.

PR #191, opened 2021-09-30
fix: Properly handle headers which aren't mangled by Rack

kgilpin:


kgilpin:
> What's the problem these changes are solving?

Sure, I've updated the description.

PR #202, opened 2021-11-03
feat: Externalize hooks config

kgilpin:
Hooks that are builtin to appmap-ruby are now all moved to YAML files in `lib/appmap/builtin_hooks` and `lib/appmap/gem_hooks`. So, builtin Ruby library and gem hooks, with labels, can now be added without writing Ruby code.

Example: https://github.com/applandinc/appmap-ruby/pull/202/files#diff-6d12bf13835e3e35920437bd8ff7f9526d256b39ab1b31c3897141e9d8b5ec31R1

```yaml
- method: OpenSSL::PKey::PKey#sign
  label: crypto.pkey
- methods:
  - OpenSSL::X509::Request#sign
  - OpenSSL::X509::Request#verify
  label: crypto.x509
```

`builtin_hooks` is for anything that's part of Ruby itself. The reason builtins are treated differently than gems is that there's no Gem path available from bundler. `gem_hooks` is for Gems. Perhaps this distinction could be removed, but I didn't want to bite that off in this PR. Probably almost all additional hooks would be gem hooks anyway.

In subsequent work, we can enable these config hooks to be added via YAML files that aren't part of `appmap-ruby` itself. For example, the `appmap-ruby` agent could be configured to look in an extension/configuration directory for additional builtin hooks.

### DIFF
--- a/lib/appmap/config.rb
+++ b/lib/appmap/config.rb
@@ -12,16 +12,17 @@ require 'appmap/depends/configuration'
 module AppMap
   class Config
     # Specifies a code +path+ to be mapped.
+    #
     # Options:
     #
     # * +gem+ may indicate a gem name that "owns" the path
-    # * +package_name+ can be used to make sure that the code is required so that it can be loaded. This is generally used with
+    # * +require_name+ can be used to make sure that the code is required so that it can be loaded. This is generally used with
     #   builtins, or when the path to be required is not automatically required when bundler requires the gem.
     # * +exclude+ can be used used to exclude sub-paths. Generally not used with +gem+.
     # * +labels+ is used to apply labels to matching code. This is really only useful when the package will be applied to
     #   specific functions, via TargetMethods.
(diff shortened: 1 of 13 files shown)


lib/appmap/config.rb
```diff
@@ -75,19 +77,16 @@ def gem_path(gem, optional)
...
           path: path,
-          require_name: require_name,
           gem: gem,
-          handler_class: handler_class.name,
+          require_name: require_name,
+          handler_class: handler_class ? handler_class.name : nil,
           exclude: Util.blank?(exclude) ? nil : exclude,
           labels: Util.blank?(labels) ? nil : labels,
-          shallow: shallow
+          shallow: shallow.nil? ? nil : shallow,
```
apotterri, line 94: This seems a little odd. Why not just `shallow: shallow` the way it was?

apotterri (approved):
This is a really nice improvement. Thanks!

PR #207, opened 2021-12-06
feat: Report sub-packages in the class map

kgilpin:
When a recorded path has recorded sub-folders, appmap-ruby will produce a class map with nested `package` elements.

### DIFF
--- a/lib/appmap/class_map.rb
+++ b/lib/appmap/class_map.rb
@@ -76,24 +76,39 @@ module AppMap
         methods.each do |method|
           add_function root, method
         end
-        root.children.map(&:to_h)
+
+        collapse_package = lambda do |package|
+          return unless package.type == 'package'
+
+          while package.children.length == 1 && package.children.all? { |child| child.type == 'package' }
+            child = package.children[0]
+            package.children.clear
+            child.children.each { |child| package.children << child }
+            package.name = [ package.name, child.name ].join('/')
+          end
+          package.tap do
+            package.children.map(&collapse_package)
+          end
+        end
+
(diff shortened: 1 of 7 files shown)


apotterri:
Do you foresee any problems if the directory structure (which will now be reflected in `package` names) doesn't match the namespacing in the source? For example, I can create a file in `foo/bar/baz.rb` that contains

```ruby
module Foo
  module Blort
    class Baz
      def m
        return 'Hello'
      end
    end
  end
end
```
and the fully-qualified name of `m` will be `Foo::Blort::Baz.m`, but its containing packages will be `foo.bar` (right?).

Asked another way: do we split class names on `::` and expect that the leading components are package names?

No matter what the answer to these questions is, do we need a test for it?

kgilpin:
> Do you foresee any problems if the directory structure (which will now be reflected in `package` names) doesn't match the namespacing in the source? For example, I can create a file in `foo/bar/baz.rb` that contains
> 
> ```ruby
> module Foo
>   module Blort
>     class Baz
>       def m
>         return 'Hello'
>       end
>     end
>   end
> end
> ```
> 
> and the fully-qualified name of `m` will be `Foo::Blort::Baz.m`, but its containing packages will be `foo.bar` (right?).
> 
> Asked another way: do we split class names on `::` and expect that the leading components are package names?
> 
> No matter what the answer to these questions is, do we need a test for it?

it's going to be reported as `foo/bar/Foo::Blort::Baz#m`. The packages are always going to be the outer enclosure, with the class hierarchy next and finally the function. We don't expect class/module names to match package/folder names.

apotterri (approved):

PR #210, opened 2022-01-11
fix: switch to activesupport's deep_dup implementation

torchello:


kgilpin:
You have to require activesupport core_ext.

Check the file history and see how it was done before.

Screen Shot 2022-01-11 at 11 39 40 AM

lib/appmap.rb
```diff
@@ -20,6 +20,13 @@
 # - appmap/swagger (Rake task)
 # - appmap/depends (Rake task)
 
+begin
+  require 'active_support'
+  require 'active_support/core_ext'
+rescue NameError
+  warn 'active_support is not available. AppMap execution will continue optimistically without it...'
```
kgilpin, line 27: I guess this is fine. 

Due to this line:

https://github.com/applandinc/appmap-ruby/blob/6145d074fa22aca7e0979510ef3118b5e32ff991/appmap.gemspec#L35

it should never fail.

kgilpin (approved):

PR #212, opened 2022-01-19
fix: Stack overflow error on prepend

kgilpin:


lib/appmap/hook/method.rb
```diff
@@ -94,7 +94,9 @@ def activate
         end
         hook_method_def = hook_method_def.ruby2_keywords if hook_method_def.respond_to?(:ruby2_keywords)
 
-        hook_class.define_method_with_arity(hook_method.name, hook_method.arity, hook_method_def)
+        m = Module.new
```
apotterri, line 97: This is going to create a new Module for every method that gets hooked, right? That seems like it could increase memory consumption by quite a bit.

Is there some way of determining that a prepend has happened previously, and only do it in that case? Why does the stack overflow happen if we don't do this?

kgilpin: Well, I don't know that the overhead is that much compared to what we're already doing? I'm not sure of another way to do this with the same certainty as this approach.

apotterri (approved):
LGTM.

PR #214, opened 2022-01-24
fix: Update labels for deserialization

kgilpin:
Deserialization comes in two flavors:

* `deserialize.safe` can deserialize untrusted data safely, because it only builds basic types and doesn't have any other weaknesses such as accessing the network or file system.
* `deserialize.unsafe` can only deserialize trusted data, because it can be configured/tricked into performing unsafe behaviors.

These labels will be used in a Rule that checks for deserialization of untrusted data.

Related: https://github.com/applandinc/scanner/pull/88


lib/appmap/gem_hooks/activerecord.yml
```diff
@@ -1,2 +1,4 @@
 - method: ActiveRecord::Relation#records
   label: dao.materialize
+- method: ActiveRecord::FixtureSet::File#raw_rows
+#  label: deserialize.safe
```
apotterri, line 4: Why are these labels commented out?

lib/appmap/builtin_hooks/yaml.yml
```diff
@@ -8,13 +8,12 @@
   - Psych#parse_stream
   label:
     - format.yaml.parse
-    - deserialize
     - deserialize.unsafe
```
dividedmind, line 11: Actually `parse` methods are safe -- they don't build objects but simply return the document tree or stream. In fact they can be removed from here -- they're not deserialization but parsing. `format.yaml.parse` seems like an adequate label (but only for these: `load` functions actually use `parse`), though I'm not sure about its utility.

Whole psych should probably be labeled something like:
```yaml
format.yaml.parse: Psych.parse_stream  # parse actually uses this one
deserialize.unsafe: Psych.load, Psych.load_stream  # or Psych::Nodes::Node#to_ruby, though this seems like overkill
deserialize.safe: Psych.safe_load
```

Not sure about

kgilpin:
Well, I don't want to leave those labels in because I'm not 100% sure that
they are correct - but I don't want to lose track of the fact that I dug
into this code to find them either.

Having these methods in the gem hooks will ensure that they are recorded.
That way, if there's a finding (
https://github.com/applandinc/scanner/pull/88), this method will be there
for context.

Eventually we should either commit these labels, or remove the method hook.

lib/appmap/gem_hooks/activerecord.yml
dividedmind: And why would they be needed anyway?

dividedmind (changes requested):

lib/appmap/gem_hooks/activerecord.yml
kgilpin: An example would be `YAML.load` - while this method is unsafe, it could be used in a context that guarantees it's actually safe. For example, Rails loads configuration with `YAML.load`. Personally, I wonder why they don't use `safe_load`. But, since the Rails configuration is subject to developer approval, a good argument can be made that the data itself is safe, and therefore even though `YAML.load` is unsafe, using YAML.load to load Rails configuration is not actually unsafe, and therefore we shouldn't report a finding for such a situation.

In general, an event labeled `deserialize.unsafe` that is a descendant of an event labeled `deserialize.safe` will not trigger a finding.

lib/appmap/builtin_hooks/yaml.yml
kgilpin: OK, I'll update it.

lib/appmap/gem_hooks/activerecord.yml
dividedmind: Thanks, that makes a lot of sense.

PR #218, opened 2022-02-10
fix: Ensure Content-Type is included in request headers

kgilpin:
Rack doesn’t prepend HTTP_ to the content type header, so we are usually/always missing it on requests.

### DIFF
--- a/lib/appmap/util.rb
+++ b/lib/appmap/util.rb
@@ -121,15 +121,21 @@ module AppMap
           blank?(headers) ? nil : headers
         end
 
-        # Rack prepends HTTP_ to all client-sent headers.
-
         if !env['rack.version']
           warn "Request headers does not contain rack.version. HTTP_ prefix is not expected."
           return finalize_headers.call(env.dup)
         end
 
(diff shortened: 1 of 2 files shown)


lib/appmap/util.rb
```diff
@@ -121,15 +121,16 @@ def select_rack_headers(env)
           blank?(headers) ? nil : headers
         end
 
-        # Rack prepends HTTP_ to all client-sent headers.
+        # Rack prepends HTTP_ to all client-sent headers (except Content-Type and Content-Length?).
```
apotterri, line 124: It seemed so bizarre to me that these two headers would be treated differently, I went off and researched why.

Turns out rack is just following the CGI spec: https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.18 .

Will you add a comment to that effect, please, and include that link to the RFC?

apotterri (changes requested):
🤯

kgilpin:
Ok. Maybe we should be sure and get Authorization also.

PR #221, opened 2022-02-16
fix: Define method signature dynamically

kgilpin:
Create method signature for the hook method that matches the argument
names and types of the original method. This ensures that the hook
method remains compatible with any code that inspects parameters
dynamically.


apotterri (approved):
LGTM!

kgilpin:
Solved via alternate means.

PR #222, opened 2022-02-16
fix: Retain the proper signature on hooked methods

kgilpin:


apotterri (approved):
Nice. This is definitely cleaner than the other approach. I like it.

PR #223, opened 2022-02-17
fix: Override method accessors to provide the correct signature

kgilpin:
Patch Object and Module to provide the proper parameters (method signature) on hooked methods.

### DIFF
--- a/lib/appmap/hook/method.rb
+++ b/lib/appmap/hook/method.rb
@@ -25,6 +25,8 @@ module AppMap
 
       ARRAY_OF_EMPTY_HASH = [{}.freeze].freeze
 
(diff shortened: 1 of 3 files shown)


lib/appmap/hook/method.rb
```diff
@@ -100,11 +102,34 @@ def activate
...
+                define_method(:parameters) do
+                  signature
+                end
+              end
+            end
+
+            method
+          end
+
+          cls.module_eval do
```
kgilpin, line 125: I don't think I need to do this over and over; just once on the base type. Right?

apotterri (approved):
Looks like these commits could be squashed?

PR #225, opened 2022-03-06
fix: Remove GC before test case execution, because it's slow

kgilpin:


lib/appmap/hook/method.rb
```diff
@@ -70,18 +70,24 @@ def activate
...
           }
 
           call_instance_method = -> {
             # https://github.com/applandinc/appmap-ruby/issues/153
             if NEW_RUBY && is_array_containing_empty_hash.(args) && hook_method.arity == 1
-              instance_method.call({}, &block)
+              if NEW_RUBY
+                hook_method.bind_call(self, {}, &block)
+              else
+                hook_method.bind(self).call({}, &block)
```
dividedmind, line 83: This line is unreachable.

PR #227, opened 2022-03-10
feat: Security labels plus optimizations

kgilpin:


lib/appmap/hook.rb
```diff
@@ -116,6 +116,11 @@ def hook_builtins
               end
 
               methods = []
+              # irb(main):001:0> Kernel.public_instance_method(:system)
+              # (irb):1:in `public_instance_method': method `system' for module `Kernel' is  private (NameError)
```
dividedmind, line 123: So weird! O_o

lib/appmap/hook/method.rb
```diff
@@ -69,28 +69,24 @@ def activate
...
-          call_instance_method = -> {
-            # https://github.com/applandinc/appmap-ruby/issues/153
-            if NEW_RUBY && is_array_containing_empty_hash.(args) && hook_method.arity == 1
-              if NEW_RUBY
-                hook_method.bind_call(self, {}, &block)
-              else
-                hook_method.bind(self).call({}, &block)
-              end
+        is_array_containing_empty_hash = ->(obj) {
+          obj.is_a?(Array) && obj.length == 1 && obj[0].is_a?(Hash) && obj[0].size == 0
```
dividedmind, line 73: I wonder, wouldn't it work to replace this method with `args == [{}]`?

lib/appmap/hook/method.rb
```diff
@@ -69,28 +69,24 @@ def activate
...
-              if NEW_RUBY
-                hook_method.bind_call(self, *args, &block)
-              else
-                hook_method.bind(self).call(*args, &block)
-              end
+              hook_method.bind(receiver).call(*args, &block)
             end
-          }
+          end
+        end
```
dividedmind, line 87: Might make sense to try to skip the NEW_RUBY branch on every call:
```ruby
if NEW_RUBY
  call_instance_method = lambda do |receiver, args, &block|
    if args == [{}] && hook_method.arity == 1
      hook_method.bind_call(receiver, {}, &block)
    else
      hook_method.bind_call(receiver, *args, block)
    end
  end
else
  call_instance_method = lambda do |receiver, args, &block|
    hook_method.bind(receiver).call(*args, &block)
  end
end
```

BTW, is this [{}] workaround still needed for ruby 3?

spec/record_net_http_spec.rb
```diff
@@ -127,7 +127,7 @@ def stop_recording
         :message:
         - :name: ary
           :class: Array
-          :value: '["1", "2"]'
+          :value: "[1, 2]"
```
dividedmind, line 130: Those stringification changes don't look good to me. Before them, the values where valid Ruby literals. after the change they're not only invalid, but also ambiguous (consider `[1, 2]` -- `[1, 2]`? `["1", 2]`? `[1, "2"]`? or `{:foo: 4}` -- `{foo: '4'}`? `{":foo" => 4}`? ...).

dividedmind (changes requested):
I'm not sure about these stringification changes.

spec/record_net_http_spec.rb
kgilpin: This should really not be language-specific though. The most likely place for the user to see this data is in the AppMap view. The hash-rocket really didn't make sense as a formatting. Maybe something closer to JSON - but overall the goal here is to make `stringify` fast, because it takes a lot of the time in AppMap.

dividedmind: > This should really not be language-specific though.

Why not? The user presumably knows the language and the principle of least surprise dictates it to be in the usual language-specific form. Using the `{a: 3}` shorthand for symbol keys is fine. `{:a: 3}` is the worst of both worlds, plus the ambiguity can be confusing and potentially important.

> but overall the goal here is to make stringify fast, because it takes a lot of the time in AppMap.

Did you measure if it's an improvement? It's not entirely clear to me that it should be (maybe except limiting the number of elements).

dividedmind: Also, it should simply limit the amount of elements displayed. With the changes, what's `[42]`? Real `[42]`? Or `['42']`? Or maybe `[42] * 42`? These all stringify to `[42]` after the changes.

lib/appmap/hook/method.rb
kgilpin: Maybe but, I'm afraid to change this.

spec/record_net_http_spec.rb
kgilpin: I've updated the stringification. Please take another look.

PR #235, opened 2022-03-28
fix: Rescue exceptions when calling Class#to_s

### LINKED ISSUE #226: Rails: Errors with relation `has_and_belongs_to_many`
Models with a `has_and_belongs_to_many` relationship are causing a strange runtime error when trying to map with Minitest:
```
Error:

RuntimeError: Wrapped undumpable exception for: ArgumentError: Class name cannot be blank. You need to supply a name argument when anonymous class given
    app/models/instance.rb:6:in `<class:Instance>'
    app/models/instance.rb:3:in `<top (required)>'
```

The model files are very basic:
```
# frozen_string_literal: true

class Instance < ApplicationRecord
  belongs_to :course
  has_and_belongs_to_many :instructor
end
```
```
# frozen_string_literal: true

class Instructor < ApplicationRecord
  belongs_to :user
  has_and_belongs_to_many :instance
end
```

Removing the relations clears up the error.

***

Ruby: 3.1.0
Rails: 7.0.2.2
AppMap: 0.73.0

apotterri:
Calling #to_s on Rails internal classes (e.g. those used for
associations) may raise an exception. Fall back to using the class name
if that happens.

These changes also add a minimal Rails 7 app for testing. This app is
complete enough to show that the #to_s fix works, but not enough to
claim that we should claim Rails 7 support.

Fixes #226.


kgilpin:
Is it worth having an ignore list of classes that are known to have bad
behaving to_s?

spec/rails_test_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails_spec_helper'
+
+# Rails5 doesn't work with Ruby 3.x
+def default_rails_versions
+  # TODO: consider testing other versions of Rails
+  # testing_ruby_2? ? [ 5, 6 ] : [ 6 ]
+  [ 7 ]
```
kgilpin, line 7: Is this something that will be changed back before merge?

apotterri: I updated this so 5, 6, 7 will all be run. The test only checks that `APPMAP=true bundle exec rake` succeeds, though, which is enough to show that the fix works. We can follow up with some more tests.

kgilpin (approved):

PR #236, opened 2022-03-29
feat: Parameter schema, query EXPLAIN and size

kgilpin:
Implement https://github.com/applandinc/appmap/pull/29

1) Hook template rendering so that we can grab the `:json` argument
2) Always emit the parameter schema of an HTTP server response; optionally emit it when enabled by APPMAP_PARAMETER_SCHEMA
3) Emit `size` of Array and Hash objects
4) Emit query `EXPLAIN` plan when enabled by APPMAP_EXPLAIN_QUERIES

### DIFF
--- a/lib/appmap/agent.rb
+++ b/lib/appmap/agent.rb
@@ -104,5 +104,9 @@ module AppMap
     def parameter_schema?
       ENV['APPMAP_PARAMETER_SCHEMA'] == 'true'
     end
+
+    def explain_queries?
+      ENV['APPMAP_EXPLAIN_QUERIES'] == 'true'
+    end
   end
 end
--- a/lib/appmap/agent.rb
+++ b/lib/appmap/agent.rb
@@ -100,5 +100,9 @@ module AppMap
       @metadata ||= Metadata.detect.freeze
       Util.deep_dup(@metadata)
     end
+
+    def parameter_schema?
+      ENV['APPMAP_PARAMETER_SCHEMA'] == 'true'
+    end
   end
 end
--- a/lib/appmap/event.rb
+++ b/lib/appmap/event.rb
@@ -60,16 +60,37 @@ module AppMap
           final ? value_string : encode_display_string(value_string)
         end
 
-        def object_properties(hash_like)
-          hash = hash_like.to_h
-          hash.keys.map do |key|
-            {
+        def add_schema(param, value, always: false)
+          return unless always || AppMap.parameter_schema?
+
+          if value.blank? || value.is_a?(String)
+            # pass
+          elsif value.is_a?(Enumerable)
+            if value.is_a?(Hash)
+              param[:properties] = object_properties(value)
+            elsif value.respond_to?(:first) && value.first
+              param[:properties] = object_properties(value.first)
+            end
+          elsif value.respond_to?(:as_json)
+            add_schema param, JSON.parse(value.to_json), always: always
+          end
+        end
+
+        def object_properties(hash)
+          hash = hash.attributes if hash.respond_to?(:attributes)
+          hash = hash.to_h if hash.is_a?(Struct)
+
+          return unless hash.respond_to?(:each_with_object)
+          
+          hash.each_with_object([]) do |entry, memo|
+            key, value = entry
+            memo << {
               name: key,
-              class: hash[key].class.name,
+              class: value.class.name,
             }
           end
         rescue
-          nil
+          warn $!
         end
 
         # Heuristic for dynamically defined class whose name can be nil
@@ -221,7 +242,9 @@ module AppMap
                 object_id: value.__id__,
                 value: display_string(value),
                 kind: param_type
-              }
+                }.tap do |param|
+                  add_schema param, value
+                end
             end
             event.receiver = {
               class: best_class_name(receiver),
@@ -276,7 +299,7 @@ module AppMap
       attr_accessor :return_value, :exceptions
 
       class << self
-        def build_from_invocation(parent_id, return_value, exception, elapsed: nil, event: MethodReturn.new)
+        def build_from_invocation(parent_id, return_value, exception, elapsed: nil, event: MethodReturn.new, parameter_schema: false)
           event ||= MethodReturn.new
           event.tap do |_|
             if return_value
@@ -284,7 +307,9 @@ module AppMap
                 class: best_class_name(return_value),
                 value: display_string(return_value),
                 object_id: return_value.__id__
-              }
+              }.tap do |param|
+                add_schema param, return_value, always: parameter_schema
+              end
             end
             if exception
               next_exception = exception
--- a/lib/appmap/event.rb
+++ b/lib/appmap/event.rb
@@ -61,6 +61,8 @@ module AppMap
         end
 
         def add_schema(param, value, always: false)
+          param[:size] = value.size if value.respond_to?(:size) && value.is_a?(Enumerable)
+
           return unless always || AppMap.parameter_schema?
 
           if value.blank? || value.is_a?(String)
@@ -71,27 +73,53 @@ module AppMap
             elsif value.respond_to?(:first) && value.first
               param[:properties] = object_properties(value.first)
             end
-          elsif value.respond_to?(:as_json)
-            add_schema param, JSON.parse(value.to_json), always: always
+          else
+            json_value = try_as_json(value) || try_to_json(value) 
+            if value != json_value
+              add_schema param, json_value, always: always
+            end
           end
         end
+
+        def try_as_json(value)
+          value.respond_to?(:as_json) && value.as_json
+        end
+        private_instance_methods :try_as_json
+
+        def try_to_json(value)
+          value.respond_to?(:to_json) && JSON.parse(value.to_json)
+        end
+        private_instance_methods :try_to_json
 
         def object_properties(hash)
           hash = hash.attributes if hash.respond_to?(:attributes)
-          hash = hash.to_h if hash.is_a?(Struct)
+
+          hash = try_to_h(hash)
 
           return unless hash.respond_to?(:each_with_object)
           
-          hash.each_with_object([]) do |entry, memo|
-            key, value = entry
-            memo << {
-              name: key,
-              class: value.class.name,
-            }
-          end
+          hash.map { |k, v| { name: k, class: v.class.name } }
         rescue
           warn $!
         end
+
+        def try_to_h(value)
+          return value unless value.respond_to?(:to_h)
+
+          # Includes such bad actors as Psych::Nodes::Scalar.
+          # Also don't try and hashifiy list-ish things.
+          @unhashifiable_classes ||= Set.new([ Array, Set ])
+
+          return value if @unhashifiable_classes.include?(value.class)
+
+          begin
+            value.to_h
+          rescue
+            # warn "#{value.class}#to_h failed: #{$!.message}"
+            @unhashifiable_classes << value.class
+          end
+        end
+        private_instance_methods :try_to_h
 
         # Heuristic for dynamically defined class whose name can be nil
         def best_class_name(value)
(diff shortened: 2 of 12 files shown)


lib/appmap/handler/rails/request_handler.rb
```diff
@@ -46,8 +49,7 @@ def to_h
                     value: self.class.display_string(val),
                     object_id: val.__id__,
                   }.tap do |message|
-                    properties = object_properties(val)
-                    message[:properties] = properties if properties
+                    AppMap::Event::MethodEvent.add_schema message, val, always: true
```
dividedmind, line 50: It appears this function is not available as of this commit. It's only introduced in the next one.

lib/appmap/agent.rb
```diff
@@ -100,5 +100,9 @@ def detect_metadata
       @metadata ||= Metadata.detect.freeze
       Util.deep_dup(@metadata)
     end
+
+    def parameter_schema?
+      ENV['APPMAP_PARAMETER_SCHEMA'] == 'true'
```
dividedmind, line 105: Perhaps this could be memoized to avoid repeated string comparisons (although, granted, most of the time it'd just be a `nil` so I suppose it's not that important).

lib/appmap/event.rb
```diff
@@ -60,16 +60,37 @@ def display_string(value)
...
+          if value.blank? || value.is_a?(String)
+            # pass
+          elsif value.is_a?(Enumerable)
+            if value.is_a?(Hash)
+              param[:properties] = object_properties(value)
+            elsif value.respond_to?(:first) && value.first
+              param[:properties] = object_properties(value.first)
+            end
+          elsif value.respond_to?(:as_json)
+            add_schema param, JSON.parse(value.to_json), always: always
```
dividedmind, line 75: Why not use `value.as_json` directly instead of doing a potentially expensive roundtrip?

lib/appmap/event.rb
```diff
@@ -60,16 +60,37 @@ def display_string(value)
...
+          end
+        end
+
+        def object_properties(hash)
+          hash = hash.attributes if hash.respond_to?(:attributes)
+          hash = hash.to_h if hash.is_a?(Struct)
```
dividedmind, line 81: Why not do `.to_h` unconditionally at this point? That way it would work with more things than just hashes and structs.

lib/appmap/event.rb
```diff
@@ -60,16 +60,37 @@ def display_string(value)
...
+
+          return unless hash.respond_to?(:each_with_object)
+          
+          hash.each_with_object([]) do |entry, memo|
```
dividedmind, line 85: You could simply do `hash.map { |k, v| { name: k, class: v.class.name } }`. This whole `memo` business is a bit difficult to follow.

lib/appmap/event.rb
```diff
@@ -221,7 +242,9 @@ def build_from_invocation(defined_class, method, receiver, arguments, event: Met
                 object_id: value.__id__,
                 value: display_string(value),
                 kind: param_type
-              }
+                }.tap do |param|
+                  add_schema param, value
+                end
```
dividedmind, line 248: This block is misaligned. Please remember to rubocop these commits :)

lib/appmap/event.rb
```diff
@@ -243,6 +243,7 @@ def build_from_invocation(defined_class, method, receiver, arguments, event: Met
                 value: display_string(value),
                 kind: param_type
                 }.tap do |param|
+                  param[:size] = value.size if value.respond_to?(:size) && value.is_a?(Enumerable)
```
dividedmind, line 246: Seems like this should be handled by `add_schema`.

lib/appmap/agent.rb
```diff
@@ -104,5 +104,9 @@ def detect_metadata
     def parameter_schema?
       ENV['APPMAP_PARAMETER_SCHEMA'] == 'true'
     end
+
+    def explain_queries?
+      ENV['APPMAP_EXPLAIN_QUERIES'] == 'true'
```
dividedmind, line 109: Ditto.

lib/appmap/handler/rails/sql_handler.rb
```diff
@@ -43,6 +43,20 @@ class << self
             def examine(payload, sql:)
               return unless (examiner = build_examiner)
 
+              if AppMap.explain_queries? && examiner.in_transaction? && examiner.database_type == :postgres
```
dividedmind, line 46: Why only do it in transaction?

lib/appmap/handler/rails/sql_handler.rb
```diff
@@ -43,6 +43,20 @@ class << self
...
+                unless sql =~ /\A(SAVEPOINT|RELEASE|ROLLBACK|BEGIN|INSERT|COMMIT)/i
```
dividedmind, line 47: That selection is somewhat haphazard. Why exclude `INSERT` but include `CREATE`, for example? I think it would make more sense to simply have a whitelist of the four query commands, ie. `SELECT`, `INSERT`, `DELETE` and `UPDATE`, plus a `WITH` for CTEs (which are only valid at the start in queries).

lib/appmap/handler/rails/sql_handler.rb
```diff
@@ -93,8 +111,12 @@ def database_type
               type
             end
 
+            def in_transaction?
+              ActiveRecord::Base.connection.open_transactions > 0
+            end
+
             def execute_query(sql)
-              ActiveRecord::Base.connection.execute(sql).inject([]) { |memo, r| memo << r; memo }
+              ActiveRecord::Base.connection.execute(sql).each_with_object([]) { |r, memo| memo << r }
```
dividedmind, line 119: Aka. `to_a`?

dividedmind (changes requested):

lib/appmap/event.rb
kgilpin: Some objects that implement `to_json` don't implement `as_json`. `as_json` is an ActiveModel thing - so it'll be common, but not required. Still, I could try `as_json` first and then fall back on the round-trip.

lib/appmap/event.rb
kgilpin: I will give it a try.

lib/appmap/event.rb
kgilpin: Unless you do it every day :-)

lib/appmap/event.rb
kgilpin: Yup.

lib/appmap/handler/rails/request_handler.rb
kgilpin: I am doing my best to separate the commits but don't always nail it completely :-)

lib/appmap/handler/rails/sql_handler.rb
kgilpin: I want to use SAVEPOINT in case EXPLAIN something blows up. SAVEPOINT is illegal outside of a transaction.

lib/appmap/handler/rails/sql_handler.rb
kgilpin: OK, sounds good.

lib/appmap/event.rb
kgilpin: There are various types that respond to `to_h` but blow up if you try and use it. I'm not sure how to address it. Here are some examples:

```
Array#to_h failed: wrong element type Set at 0 (expected array)
Set#to_h failed: wrong element type String (expected array)
Psych::Nodes::Scalar#to_h failed: wrong element type Psych::Nodes::Scalar (expected array)
```

lib/appmap/event.rb
dividedmind: But you're looking for `as_json` in the condition, so all objects will have it at this point. Unless that's not what you wanted :)

lib/appmap/event.rb
dividedmind: It's very un-rubyish ;)

lib/appmap/handler/rails/sql_handler.rb
dividedmind: But it'd still be nice to EXPLAIN even outside of a transaction.

lib/appmap/event.rb
```diff
@@ -71,28 +73,54 @@ def add_schema(param, value, always: false)
...
         end
 
+        def try_to_h(value)
+          return value unless value.respond_to?(:to_h)
+
+          # Includes such bad actors as Psych::Nodes::Scalar.
+          # Also don't try and hashifiy list-ish things.
+          @unhashifiable_classes ||= Set.new([ Array, Set ])
+
+          return value if @unhashifiable_classes.include?(value.class)
```
dividedmind, line 113: You can also check if `to_h` is generic, ie. `value.method(:to_h).owner == Enumerable`. (Array has its own implementation which does the same thing apparently, so you still have to check for this case.)

lib/appmap/event.rb
```diff
@@ -71,28 +73,54 @@ def add_schema(param, value, always: false)
...
+
+          begin
+            value.to_h
+          rescue
```
dividedmind, line 117: Rubocop! Add specific exception here, ie. `TypeError`. Or squash the rubocop warning.

lib/appmap/event.rb
```diff
@@ -71,28 +73,54 @@ def add_schema(param, value, always: false)
...
+            # warn "#{value.class}#to_h failed: #{$!.message}"
+            @unhashifiable_classes << value.class
+          end
+        end
+        private_instance_methods :try_to_h
```
dividedmind, line 122: I don't think `private_instance_methods` does what you think it does ;)

It generally works better to just put `private` before the private section of the class.

lib/appmap/event.rb
```diff
@@ -276,15 +327,18 @@ class MethodReturn < MethodReturnIgnoreValue
...
           event ||= MethodReturn.new
           event.tap do |_|
             if return_value
               event.return_value = {
                 class: best_class_name(return_value),
                 value: display_string(return_value),
                 object_id: return_value.__id__
-              }
+              }.tap do |param|
+                param[:size] = return_value.size if return_value.respond_to?(:size) && return_value.is_a?(Enumerable)
```
dividedmind, line 339: This is now handled by `add_schema` so you can remove this line. I asked for this precisely because of this duplication :)

lib/appmap/handler/rails/sql_handler.rb
```diff
@@ -43,6 +43,20 @@ class << self
...
+              if AppMap.explain_queries? && examiner.in_transaction? && examiner.database_type == :postgres
+                if sql =~ /\A(SELECT|INSERT|DELETE|UPDATE|WITH)/i
+                  examiner.execute_query 'SAVEPOINT appmap_sql_examiner'
+                  begin
+                    plan = examiner.execute_query(%(EXPLAIN #{sql}))
+                    payload[:query_plan] = plan.map { |line| line[:'QUERY PLAN'] }.join("\n")
+                    examiner.execute_query 'RELEASE SAVEPOINT appmap_sql_examiner'
+                  rescue
+                    warn "Exception occurred explaining query: #{$!}"
+                    examiner.execute_query 'ROLLBACK TO SAVEPOINT appmap_sql_examiner'
```
dividedmind, line 55: You can probably unconditionally rollback to savepoint.

dividedmind (changes requested):
Thanks! Can we please add test cases for all these object schema heuristics?

lib/appmap/event.rb
kgilpin: I have no idea what the error would be, I just don't want to blow up someone's app because appmap encounters an issue here.

kgilpin: I don't really run rubocop.

lib/appmap/event.rb
kgilpin: This line and schema are not that closely interrelated in the new iteration.

PR #239, opened 2022-03-31
Hook Kernel#eval

dividedmind:
This builds on #238 to implement changes sought by #228 .


kgilpin:
Ok I’m quite interested to take a look at this :-) Maybe @apotterri too?

lib/appmap/handler/eval.rb
```diff
@@ -0,0 +1,41 @@
...
+    # To avoid this, we explicitly substitute the correct execution
+    # context, up several stack frames.
+    class Eval < Function
+      # The depth of the frame we need to pluck out:
+      # 1. Hook::Method#do_call
+      # 2. Hook::Method#trace_call
+      # 3. Hook::Method#call
+      # 4. proc generated by Hook::Method#hook_method_def
+      # 5. the (intended) frame of the original eval that we hooked
+      # Note it needs to be adjusted if this call sequence changes.
```
kgilpin, line 32: Is there a test case that will warn us if/when this happens?

lib/appmap/handler/eval.rb
```diff
@@ -0,0 +1,41 @@
...
+    # If the binding is not provided, by default eval will run in the
+    # current frame. Since we call it here, this will mean the #do_call
+    # frame, which would make AppMap::Handler::Eval the lexical scope
+    # for constant lookup and definition; as a consequence
+    # eg. `eval "class Foo; end"` would define
+    # AppMap::Handler::Eval::Foo instead of defining it in
+    # the module where the original call was made.
+    #
+    # To avoid this, we explicitly substitute the correct execution
+    # context, up several stack frames.
```
kgilpin, line 24: Nice description, thanks. I understand the problem and the solution from this.

lib/appmap/handler/eval.rb
dividedmind: Yes, the test in `spec/handler/eval_spec.rb` will fail if the depth is not exactly right.

apotterri (approved):
👍

PR #241, opened 2022-04-03
fix: Query server_version when connection is made

apotterri:
**DO NOT MERGE THIS**  until [f8461bb](https://github.com/applandinc/appmap-ruby/pull/241/commits/f8461bb0285a3ce1bce12f3f91bd245a66159489) has been removed.

Add a Sequel::Database.after_initialize hook to cache server_version
when a connection gets initialized. This ensures that we won't try to
execute a query to get it while Sequel is executing another query.

### DIFF
--- a/lib/appmap/handler/rails/request_handler.rb
+++ b/lib/appmap/handler/rails/request_handler.rb
@@ -100,15 +100,14 @@ module AppMap
 
           protected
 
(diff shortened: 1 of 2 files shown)


kgilpin:
This does not seem to be happening in the wild.

PR #243, opened 2022-04-08
fix: Don't call #size on complex objects

kgilpin:
Limit invocation of #size to known-safe objects - initially, Array and
Hash.


apotterri (approved):
👍

PR #245, opened 2022-04-26
fix: Symbol in request env

kgilpin:


lib/appmap/class_map.rb
```diff
@@ -78,7 +78,7 @@ def build_from_methods(methods)
         end
 
         collapse_package = lambda do |package|
-          return unless package.type == 'package'
+          next unless package.type == 'package'
```
kgilpin, line 81: `return` here works on Ruby 2.6+, but not on 2.5. Other occurrences below.

lib/appmap/handler/rails/template.rb
```diff
@@ -147,7 +147,7 @@ def path_from_obj(path_obj)
               else
                 path_obj.inspect
               end
-            path = path[Dir.pwd.length + 1..] if path.index(Dir.pwd) == 0
+            path = path[Dir.pwd.length + 1..-1] if path.index(Dir.pwd) == 0
```
kgilpin, line 150: The end index is required for Ruby 2.5.

lib/appmap/hook/method.rb
```diff
@@ -51,18 +53,31 @@ def activate
         if hook_class == Kernel
           hook_class.define_method_with_arity(hook_method.name, hook_method.arity, hook_method_def)
         else
-          hook_class.ancestors.find { |cls| cls.method_defined?(hook_method.name, false) }.tap do |cls|
```
kgilpin, line 54: The second argument here (`false`) is not available in Ruby 2.5.

lib/appmap/util.rb
```diff
@@ -122,7 +122,7 @@ def select_rack_headers(env)
         # Apparently, it's following the CGI spec in doing so.
         # https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.18
         matching_headers = env
-          .select { |k,v| k.start_with? 'HTTP_' }
+          .select { |k,v| k.to_s.start_with? 'HTTP_' }
```
kgilpin, line 125: This bug would blow up the agent when a third party gem would put symbols in the request env. (datadog).

PR #247, opened 2022-04-27
fix: Install code path

kgilpin:
Rails source path can now be simply `app`.

Also emit `language` and `appmap_dir` to *appmap.yml*.


apotterri (approved):
LGTM

kgilpin:
@dustinbyrne FYI it would make life easier for our CLI tools if the appmap.yml indicated what the language is and where the appmap files are located. This PR includes that for Ruby.

apotterri:
> @dustinbyrne FYI it would make life easier for our CLI tools if the appmap.yml indicated what the language is and where the appmap files are located. This PR includes that for Ruby.

I'll include this change for the Java agent in my next set of updates (for labeling).

PR #249, opened 2022-05-05
fix: Allow appmap_dir, language and additional properties

kgilpin:


config-schema.yml
```diff
@@ -1,10 +1,14 @@
 type: object
-additionalProperties: false
+additionalProperties: true
```
apotterri, line 2: It seems like we should set `additionalProperties` to `true` everywhere. 

WDYT?

apotterri (changes requested):

kgilpin:
I think at the top level, definitely, so the AppMap config can be extended.
But within sections like packages and functions, it makes sense to be
strict so that we catch spelling errors and stuff like that.

PR #257, opened 2022-06-07
Adding version requirements to README

brikelly:
These are listed on our docs site but should be here too in case someone comes in via GitHub.


kgilpin:
Actually, no, we don't want to duplicate this information. It gets out of date and becomes hard to maintain (think, 4x clients, 2x extension platforms). We can add a link to the appmap-ruby docs on our website.

PR #259, opened 2022-06-22
fix: Exclude AppMaps from the gem when packaging

apotterri:
If there are AppMaps in the gem, the VS Code extension's Install Guide sees them and thinks recordings have been made.


kgilpin (approved):

kgilpin:
@apotterri I guess it's only an issue if the gem is vendored?

kgilpin:
OK, if you put .direnv in your `Find: Exclude` and `Watch: Exclude`
settings you'll get the same result.

PR #266, opened 2022-08-09
feat: Don't record rspec tests when appmap: false

kgilpin:


lib/appmap/rspec.rb
```diff
@@ -157,14 +157,20 @@ def begin_spec(example)
...
+
+        @recordings_by_example[example.object_id] = recording
       end
 
       def end_spec(example, exception:)
         recording = @recordings_by_example.delete(example.object_id)
         return warn "No recording found for #{example}" unless recording
 
-        recording.finish exception
+        recording.finish exception unless recording === :false
```
apotterri, line 173: Did you mean to use `===` here?

apotterri (approved):
One comment, but looks good otherwise.

PR #267, opened 2022-08-10
feat: Add builtin labels for JWT

dustinbyrne:
For use with https://github.com/applandinc/appmap-js/pull/703


lib/appmap/builtin_hooks/jwt.yml
```diff
@@ -0,0 +1,11 @@
+- method: JWT#decode
+  label: jwt.decode
```
kgilpin, line 2: FYI this should be in gem_hooks unless jwt is part of the Ruby distribution

PR #271, opened 2022-08-23
perf: Implement `custom_display_string` in a C extension

symwell:
Optimize `custom_display_string` by implementing most of it in a C extension.
- Strings are converted to unicode in Ruby.  Converting them in C improves performance by 600%-900% but gives the error:
```
in `generate': source sequence is illegal/malformed utf-8 (JSON::GeneratorError)
```
- Stringifying hash tables initially lead to buffer overflows.  If it ends up being unstable it's possible to revert to a version that stringifies in Ruby.
- <del>The `.so` file built for the C extension isn't available [to the testcases](https://app.travis-ci.com/github/applandinc/appmap-ruby/jobs/580666054).  Should install it to the right directory.</del> This is fixed now.
- Strings are currently copied twice: once in our buffer, then passed to Ruby.  There should be a way to copy them only once.  Maybe be calling `rb_str_resize` and then writing in the object's string buffer.

(1) `sample_app_6th_ed`.  Improves performance by `0.084s` (2.7%)

```
# with custom_display_string running the Ruby version
APPMAP=true DISABLE_SPRING=true bundle exec rails test
Finished in 3.14949s
67 tests, 282 assertions, 0 failures, 0 errors, 0 skips

# with custom_display_string running the C version
APPMAP=true DISABLE_SPRING=true bundle exec rails test
Finished in 3.06511s
67 tests, 282 assertions, 0 failures, 0 errors, 0 skips

# with Appmap disabled
APPMAP=false DISABLE_SPRING=true bundle exec rails test
Finished in 1.98641s
67 tests, 282 assertions, 0 failures, 0 errors, 0 skips
```

(2) `solidus`.  Improves performance by `0.950s` (2.3%)

```
# Ruby version
Finished in 42.13 seconds (files took 2.91 seconds to load)
616 examples, 216 failures, 1 pending

# C version
Finished in 41.18 seconds (files took 2.87 seconds to load)
616 examples, 216 failures, 1 pending

# AppMap disabled
Finished in 32.78 seconds (files took 5.31 seconds to load)
616 examples, 216 failures, 1 pendin
```

(3) `appmap-server`.  Improves performance by `8.9s` (3.3%) compared to Ruby.  1.4% faster compared to no AppMap.

```
# Ruby version
Finished in 4 minutes 26.5 seconds (files took 2.61 seconds to load)
549 examples, 1 failure

# C extension
Finished in 4 minutes 17.6 seconds (files took 2.65 seconds to load)
549 examples, 0 failures

# AppMap disabled
Finished in 4 minutes 21.4 seconds (files took 2.65 seconds to load)
549 examples, 0 failures
```

(4) `discourse`. Improves performance by `19s` (2.4%) compared to Ruby.

```
# Ruby version
APPMAP=true bundle exec rspec spec/requests
Finished in 13 minutes 19 seconds (files took 14.84 seconds to load)
3364 examples, 40 failures, 2 pending

# C version
APPMAP=true bundle exec rspec spec/requests
Finished in 13 minutes 0 seconds (files took 14.9 seconds to load)
3364 examples, 40 failures, 2 pending

# AppMap disabled
APPMAP=false bundle exec rspec spec/requests
Finished in 4 minutes 11.5 seconds (files took 3.05 seconds to load)
3364 examples, 11 failures, 2 pending
```

(5) `bench.sh`

```
Performance difference for 100000 iterations (higher is better)
    +------------------------------------+------------+------------------+---------------+
    | symbol                             | Ruby .to_s | ruby_custom_to_s | c_custom_to_s |
    +====================================+============+==================+===============+
    | data_nil                           | 1          | 0.4              | 0.54          |
    | data_true                          | 1          | 0.22             | 0.49          |
    | data_false                         | 1          | 0.19             | 0.47          |
    | data_int                           | 1          | 0.32             | 0.63          |
    | data_float                         | 1          | 0.71             | 1.5           |
    | data_sym                           | 1          | 0.17             | 0.69          |
    | data_str_8                         | 1          | 0.03             | 0.04          |
    | data_str_16                        | 1          | 0.03             | 0.04          |
    | data_str_32                        | 1          | 0.03             | 0.04          |
    | data_str_64                        | 1          | 0.03             | 0.04          |
    | data_str_128                       | 1          | 0.02             | 0.03          |
    | data_str_above_max_string_length   | 1          | 0.02             | 0.03          |
    | data_hash                          | 1          | 0.23             | 0.22          |
    | data_hash_more_than_max            | 1          | 1.55             | 0.6           |
    | data_array_int                     | 1          | 0.28             | 0.15          |
    | data_array_int_more_than_max       | 1          | 0.43             | 0.18          |
    | data_array_float                   | 1          | 0.52             | 0.39          |
    | data_array_float_more_than_max     | 1          | 0.87             | 0.54          |
    | data_array_str8                    | 1          | 0.18             | 0.3           |
    | data_array_str8_more_than_max      | 1          | 0.26             | 0.38          |
    | data_array_mix                     | 1          | 0.31             | 0.25          |
    | data_array_hash                    | 1          | 0.24             | 0.21          |
    | data_array_hash_more_than_max      | 1          | 0.35             | 0.31          |
    | data_array_hash_hash_more_than_max | 1          | 2.28             | 0.86          |
    +------------------------------------+------------+------------------+---------------+

```


native_c_extension/c_custom_to_s.c
```diff
@@ -0,0 +1,493 @@
...
+    break;
+  }
+  case T_STRING: {
+    VALUE element_to_s = method_c_custom_to_s(self, element);
+    int string_len = RSTRING_LEN(element_to_s);
+    // +2 for the two "s. +1 for NULL
+    method_c_custom_to_s_check_buffer_size(*offset, string_len + 3, buffer_max);
+    if (quoted) {
+      snprintf(&buffer[*offset], 1 + 1, "%s", "\"");
+      *offset += 1;
```
dividedmind, line 46: Why use snprintf here? Just do `buffer[(*offset)++] = '"'`.

native_c_extension/c_custom_to_s.c
```diff
@@ -0,0 +1,493 @@
...
+    }
+    // +1 for NULL
+    snprintf(&buffer[*offset], string_len + 1, "%s", StringValuePtr(element_to_s));
```
dividedmind, line 49: Again, snprintf is unnecessary here. What's worse, I think it can do the wrong thing -- ruby strings aren't guaranteed to be null-terminated. You need to do something like
```c
memcpy(buffer + *offset, StringValuePtr(element_to_s), string_len);
*offset += string_len;
```

native_c_extension/c_custom_to_s.c
```diff
@@ -0,0 +1,493 @@
...
+    *offset += string_len;
+    if (quoted) {
+      snprintf(&buffer[*offset], 1 + 1, "%s", "\"");
+      *offset += 1;
```
dividedmind, line 53: As above.

native_c_extension/c_custom_to_s.c
```diff
@@ -0,0 +1,493 @@
...
+    return num_bytes;
+}
+
+/*
+ * Encodes the given character's Unicode code point using UTF-8 coding.
+ * Parameters:
+ *      character   Unicode code point as an unsigned integer.
+ *      utf8        Array for UTF-8 encoded bytes (can be 1..4 bytes).
+ */
+int utf8_encode(unsigned int character, unsigned char utf8[])
```
dividedmind, line 301: I don't understand why you're dealing with encoding at all. Encoding is hard and it will be subtly buggy unless you have years of polishing and bugfixing as I assume Ruby's library has.

Can't you just copy the string and imbue it with the source encoding on creation?

native_c_extension/c_custom_to_s.c
```diff
@@ -0,0 +1,493 @@
...
+    /*   counter += utf8_encode(character, (unsigned char *) &buffer_utf8_encoded[counter]); */
+    /* } */
+    /* ret = rb_utf8_str_new(buffer_utf8_encoded, counter); */
+
+    // this is 600%-900%x faster than calling Ruby but reports the error:
+    // in `generate': source sequence is illegal/malformed utf-8 (JSON::GeneratorError)
+    // ret = rb_utf8_str_new(buffer, offset);
+
+    // call Ruby function to utf8 encode instead of encode in C
+    VALUE string_unencoded = rb_str_new_cstr(buffer);
```
dividedmind, line 468: Ruby strings can have nulls inside them and one of them could have ended in the buffer, so this won't always work.

dividedmind (commented):
Hi, I took a quick look and left a couple of comments. I don't think it makes sense dealing with encoding at all, it's unnecessary headache. Re segfaults, have you run the code under valgrind or something similar at all? I saw a couple of suspicious spots, this could help you ferret them out.

native_c_extension/c_custom_to_s.c
symwell: I'm dealing with encoding because the existing `custom_display_string` encodes in utf-8. https://github.com/applandinc/appmap-ruby/blob/master/lib/appmap/event.rb#L104.  If strings don't get encoded I see errors like
```
 in `generate': source sequence is illegal/malformed utf-8 (JSON::GeneratorError)
```

I'm dealing with encoding in C because it's 6x-9x faster than encoding in Ruby.  But I didn't get it to work.

native_c_extension/c_custom_to_s.c
symwell: snprintf also copied the NULL at the end of the string.
I was hoping for a single, consistent way of building strings.

Without copying the NULL for `]` I see
```
[10, 20, 30, 40]
[10, 20, 30, 40]�g�V
```

I'll set both the char and the NULL with a macro.

native_c_extension/c_custom_to_s.c
symwell: Switched snprintf to memcpy.

native_c_extension/c_custom_to_s.c
symwell: Switched to `rb_str_new` and passing the buffer size.

symwell:
> Re segfaults, have you run the code under valgrind or something similar at all?

Running ruby with valgrind doesn't seem to work right. When I manually induce memory errors Valgrind doesn't report them.  It's as if when ruby exits valgrind exits.

Testing as standalone C binary doesn't seem to work. I think the function isn't being called right.  It crashes the VM instantly when stringifying a simple string.

native_c_extension/c_custom_to_s.c
dividedmind: You don't need the null if you keep the length, and besides, you'd only need to add it when you're done. And Ruby strings can contain nulls, so you can't rely on them anyway.

native_c_extension/c_custom_to_s.c
dividedmind: But can't you just imbue the string with the original string encoding (presumably utf8) and let ruby deal with it? I suppose it would be problematic if you have multiple different encodings going to the same output string, but this should rarely happen (if ever).

dividedmind:
> > Re segfaults, have you run the code under valgrind or something similar at all?
> 
> Running ruby with valgrind doesn't seem to work right. When I manually induce memory errors Valgrind doesn't report them. It's as if when ruby exits valgrind exits.
> 
> Testing as standalone C binary doesn't seem to work. I think the function isn't being called right. It crashes the VM instantly when stringifying a simple string.

Would you like me to take a look at that? I used to use valgrind extensively ages ago, perhaps I could make this work.

dividedmind:
Those test benchmarks don't seem encouraging; I wonder if it's worth the trouble. Especially the last one is concerning -- is it really faster with instrumentation? Makes the whole improvement look like statistically insignificant. Did you try benchmarking a larger test suite so any effect would be more pronounced?

native_c_extension/c_custom_to_s.c
symwell: Now setting the null only when done.

symwell:
> Would you like me to take a look at that? I used to use valgrind extensively ages ago, perhaps I could make this work.

Sure, why not.

native_c_extension/c_custom_to_s.c
symwell: It should be possible to reuse the encoding of the initial string.  I didn't try this. I'll try.

symwell:
> is it really faster with instrumentation

It's not that it's faster with instrumentation, it's that collecting measurements for it is unreliable. The test and machine are too noisy.  Ideally measurements should be collected on a machine with nothing else running, no desktop started, no background services, logging in from a text terminal, and get the median of some number of runs.

The test-suite with the most visible performance overhead was `sample_app_6th_ed`.

dividedmind:
> > is it really faster with instrumentation
> 
> It's not that it's faster with instrumentation, it's that collecting measurements for it is unreliable. The test and machine are too noisy. Ideally measurements should be collected on a machine with nothing else running, no desktop started, no background services, logging in from a text terminal, and get the median of some number of runs.

That's exactly my point :) We really need to run the benchmark on a larger test suite (several minutes maybe) to see the impact. Otherwise the numbers are pretty meaningless

symwell:
We decided the performance improvement of ~3% is too low for the risk it introduces.  Won't ship this.

PR #273, opened 2022-09-02
feat: Write an AppMap for each observed HTTP request

kgilpin:
https://www.loom.com/share/8a70b15f22c8496a9e61673afbb7055d

### DIFF
--- a/lib/appmap/middleware/remote_recording.rb
+++ b/lib/appmap/middleware/remote_recording.rb
@@ -92,7 +92,7 @@ module AppMap
           if tracer
             AppMap.tracing.delete(tracer)
 
-            events = tracer.events.dup.map(&:to_h)
+            events = tracer.events.map(&:to_h)
 
             appmap_name = "#{req.request_method} #{req.path} (#{status}) - #{start_time.strftime('%T.%L')}"
             appmap_file_name = AppMap::Util.scenario_filename([ start_time.to_f, req.url ].join('_'))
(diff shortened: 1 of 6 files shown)


lib/appmap/middleware/remote_recording.rb
```diff
@@ -75,10 +76,50 @@ def stop_recording(req)
...
+        # into a separate Tracer.
+        tracer = AppMap.tracing.trace(thread: Thread.current) if record_all_requests?
+
+        @app.call(env).tap do |status, headers|
+          if tracer
+            AppMap.tracing.delete(tracer)
+
+            events = tracer.events.dup.map(&:to_h)
+
+            appmap_name = "#{req.request_method} #{req.path} (#{status}) - #{start_time.strftime('%T.%L')}"
```
dustinbyrne, line 97: Adding request method, path and status to the metadata would also be useful for displaying this in an IDE without needing to parse `events`.

Some other ideas:
- Elapsed time of the request
- Query strings and elapsed times
- Exception details (stack trace?) if an exception failed the request

kgilpin: I think there are a couple of efficient strategies we can use to implement this.

After indexing, the AppMaps list `events` first. The `http_server_request` is the first event. So, a streaming JSON parser can efficiently pick out the first event, which will be the request, and analyze it for query string, elapsed time, etc. 

The index contains `canonical.httpServerRequests` (which will contain one route in this case). The `status` code is in that file - so it's easy to quickly find out if the request succeeded or failed.

Having `exceptions.json` in the index seems like it would be a nice addition. We can handle this with an enhancement to the indexer.

dustinbyrne (approved):

kgilpin:
Rather than duplicate all that in the metadata, it would be better to
enhance the indexer

PR #278, opened 2022-09-09
Sw/include method source in appmap

kgilpin:
We don't need the code to get the source code of the function - and it's also horribly inefficient. So, you can remove that along with its test cases.


symwell:
I believe this pr is superseded by https://github.com/applandinc/appmap-ruby/pull/279 which includes only the `method_source` source code AppMap uses.

kgilpin:
If so, please close it.

symwell:
Closing.

PR #281, opened 2022-09-13
feat: Record by default

kgilpin:


apotterri (approved):
Looks good to me! I miss Ruby.... ;)

PR #282, opened 2022-09-14
feat: Save default appmap.yml

kgilpin:
If the specified appmap.yml does not exist, use a default one (this is existing behavior). Also, print the default config YAML to the specified location - this is new behavior.

Screen Shot 2022-09-14 at 2 23 00 PM


apotterri (approved):
LGTM.

PR #279, opened 2022-09-09
perf: Use AppMap's optimized `extract_last_comment`

symwell:
Use our optimized version of `extract_last_comment` that runs 65x faster until the [corresponding pr](https://github.com/banister/method_source/pull/78) in `method_source` gets merged.

Fixes https://github.com/applandinc/appmap-ruby/issues/276

PR #283, opened 2022-09-16
fix: Faster comment extraction

### LINKED ISSUE #276: Take the `method_source` code we need and copy it in AppMap, to not depend on `method_source` or conflict with it in the wild
We want to take the `method_source` code we need and add it to our repo e.g. `lib/appmap/method_source`.

Then we can apply the [performance fix](https://github.com/banister/method_source/pull/78) we want in our repo. That way we can be sure not to conflict with method_source in the wild, and we will get the efficient implementation that we want.

Give proper attribution and then we can be done with it.

https://applanding.slack.com/archives/CQ5JFE1DY/p1662647402949409

dividedmind:
AppMap needs to extract method comments to apply labels. This change makes this process faster.

Note this is roughly the same change as in #279 but (IMO) a bit cleaner, plus it avoids reading the whole file by using IO#lines iterator and only taking the required number of lines. As such it fixes #276 (in a way).

### DIFF
--- a/lib/appmap/trace.rb
+++ b/lib/appmap/trace.rb
@@ -21,8 +21,11 @@ module AppMap
       end
 
       def comment
-        @method.comment
-      rescue MethodSource::SourceNotFoundError, Errno::EINVAL
+        return nil if source_location.nil? || source_location.first.start_with?('<')
+
+        # Do not use method_source's comment method because it's slow
+        @comment ||= RubyMethod.last_comment *source_location
+      rescue Errno::EINVAL, Errno::ENOENT
         nil
       end
 
@@ -36,6 +39,21 @@ module AppMap
 
       def labels
         @package.labels
+      end
+
+      private
+
+      # Read file and get last comment before line.
+      def self.last_comment(file, line_number)
+        File.open(file) do |f|
+          buffer = []
+          f.each_line.lazy.take(line_number - 1).reverse_each do |line|
+            break unless (line =~ /^\s*#/) || (line =~ /^\s*$/)
+
+            buffer << line.lstrip
+          end
+          buffer.reverse.join
+        end
       end
     end
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,7 +18,7 @@ Layout/LineLength:
   Max: 120
 
 Metrics/BlockLength:
-  ExcludedMethods:
+  AllowedMethods:
   - describe
   - it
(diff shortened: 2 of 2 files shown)


lib/appmap/trace.rb
```diff
@@ -37,6 +40,21 @@ def name
       def labels
         @package.labels
       end
+
+      private
+
+      # Read file and get last comment before line.
+      def self.last_comment(file, line)
+        File.open(file) do |f|
```
symwell, line 48: The other implementation cached file contents.  Adding caching here might be only a few lines.

symwell: `line` is used twice, as an input parameter and do parameter.

dividedmind: The other implementation cached the whole file per method, which is a bit of a waste of memory. This one caches just the comment.

symwell: I tried to test this with `method_source` testcases and got

```
$ ruby --version
ruby 3.0.1p64 (2021-04-05 revision 0fb782ee38) [x86_64-linux]

$bundle exec rspec spec

  1) MethodSource Methods should return a comment for method
     Failure/Error:
             f.lines.lazy.take(line_number - 1).reverse_each do |line|
               break unless (line =~ /^\s*#/) || (line =~ /^\s*$/)
       
               buffer << line.lstrip
             end
     
     NoMethodError:
       undefined method `lines' for #<File:/home/test/src/method_source/spec/spec_helper.rb (closed)>
       Did you mean?  lineno
     # ./lib/method_source.rb:52:in `block in comment_describing_custom'
     # ./lib/method_source.rb:49:in `open'
     # ./lib/method_source.rb:49:in `comment_describing_custom'
     # ./lib/method_source.rb:44:in `comment_helper'
     # ./lib/method_source.rb:143:in `comment'
     # ./spec/method_source_spec.rb:69:in `block (3 levels) in <top (required)>'
```

Maybe `f.lines` doesn't exist in Ruby 3.0?

Which version of Ruby did you use?

symwell: In ruby 2.7.2 and ruby 3.0.1 it works with `f.each_line`.

dividedmind:
Thanks @symwell, good catch! I've merged your changes into my commit, please take a look.

symwell (approved):
Verified locally that it works.

PR #284, opened 2022-09-19
feat: Emit metadata.recorder.type

kgilpin:
Supports AppMap version 1.9.0

https://github.com/applandinc/appmap/pull/32

### DIFF
--- a/lib/appmap/version.rb
+++ b/lib/appmap/version.rb
@@ -5,7 +5,7 @@ module AppMap
 
   VERSION = '0.90.1'
 
-  APPMAP_FORMAT_VERSION = '1.7.0'
+  APPMAP_FORMAT_VERSION = '1.9.0'
 
   SUPPORTED_RUBY_VERSIONS = %w[2.5 2.6 2.7 3.0 3.1].freeze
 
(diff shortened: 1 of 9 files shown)


lib/appmap/version.rb
```diff
@@ -5,7 +5,7 @@ module AppMap
 
   VERSION = '0.90.1'
 
-  APPMAP_FORMAT_VERSION = '1.7.0'
+  APPMAP_FORMAT_VERSION = '1.9.0'
 
   SUPPORTED_RUBY_VERSIONS = %w[2.5 2.6 2.7 3.0 3.1].freeze
```
symwell, line 10: Wondering if AppMap is deprecating support for Ruby 2.5 after seeing [this](https://github.com/applandinc/appmap-ruby/pull/282/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485L8) change in a different pr.  Doesn't need to happen in this pr.

symwell (approved):

kgilpin:
Good point. What's going on here is that 2.5 is generally known to work,
but we won't support it with are version-specific specific bug fixes, since
it's deprecated by the Ruby dev team

PR #285, opened 2022-09-21
fix: Drop database server version

kgilpin:
It's unsafe to query the DB for the version when already inside of a hook.

I experimented with various ways of querying the DB outside of hooks, but they are all super-hacky and I don't think it's worth the effort since there is nothing downstream that depends on the `server_version` being present. This field is optional in the spec.

### DIFF
--- a/lib/appmap/handler/rails/sql_handler.rb
+++ b/lib/appmap/handler/rails/sql_handler.rb
@@ -75,7 +75,6 @@ module AppMap
                 end
               end
 
-              payload[:server_version] = examiner.server_version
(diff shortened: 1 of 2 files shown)


symwell:
Can this bug be reproduced by a test?  I tried to add `payload[:server_version] = examiner.server_version` back to see if a test would break but all tests passed.

kgilpin:
I’m sure it could be, but it’s not caught by our existing test suite -
otherwise we never would have shipped it :-)

symwell (approved):

PR #293, opened 2022-10-20
fix: Don't check for rails anymore in appmap-agent-validate

### LINKED ISSUE #292: `rails` requirement is removed from `appmap-agent-validate`
The project picker only requires a web framework or a test framework be supported.

dividedmind:
Fixes #292.

Note this simply reverts #174 — neither that pull request nor the commit provides any rationale for it, but we might want to try to remember it before we merge this. 

One reason that comes to mind might have been because there is no automatic setup for non-rails project (AFAIK); the user has to know to go add `require 'appmap'` to appropriate test environment files (eg. conventionally `spec_helper.rb` for rspec) before AppMap can work. With Rails it just works because Rails environment setup autoloads all the gems from the bundle.

### DIFF
--- a/lib/appmap/service/validator/config_validator.rb
+++ b/lib/appmap/service/validator/config_validator.rb
@@ -21,7 +21,6 @@ module AppMap
         def valid?
           @violations = []
           validate_ruby_version
-          validate_rails_presence
           validate_config_presence
           parse_config
           validate_config_load
@@ -65,14 +64,6 @@ module AppMap
             @violations << Violation.error(
               filename: @config_file,
               message: "AppMap configuration #{@config_file} file does not exist"
-            )
-          end
-        end
-
-        def validate_rails_presence
-          unless Gem.loaded_specs.has_key?('rails')
-            @violations << Violation.error(
-              message: 'AppMap auto-configuration is currently not available for non Rails projects'
             )
           end
         end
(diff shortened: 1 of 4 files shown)


apotterri (approved):
I don't see any reason not to do this. I suspect it Alex did it originally just for the sake of completeness.

Looks like there's another test that needs to be fixed before it gets merged, but then it should be good to go.

dividedmind:
> I don't see any reason not to do this. I suspect it Alex did it originally just for the sake of completeness.
> 
> Looks like there's another test that needs to be fixed before it gets merged, but then it should be good to go.

What do you think about the fact that it won't work out of the box? The user needs to do some extra setup (add `require 'appmap'` somewhere, in the test helper probably) to record tests and there's currently no guidance on how to do that nor is it done automatically by `npx appmap install`. Won't we be setting the user up for silent failure?

dustinbyrne:
I think a nice compromise would be to alter the installation instructions only if Rails isn't found. E.g., make sure `require 'appmap'` is included in your entry point if it's not included via `Bundler.require` elsewhere.

At least this way we can deliver the correct information to those who need it without cluttering up the instructions for everyone else.

dividedmind:
> I think a nice compromise would be to alter the installation instructions only if Rails isn't found. E.g., make sure `require 'appmap'` is included in your entry point if it's not included via `Bundler.require` elsewhere.
> 
> At least this way we can deliver the correct information to those who need it without cluttering up the instructions for everyone else.

Yeah, that'd be nice. Unfortunately there doesn't seem to be a way for the agent to display arbitrary messages in the current framework, so it's not as simple as it could be.

dividedmind:
Superseded by #294

PR #314, opened 2023-02-02
feat: Implement `items` Parameter object format (AppMap v1.10.0)

dustinbyrne:
`items` must be captured for array-like data structures to describe the schema of the data held within. Previously array elements were assumed to be objects.

This means we should now be able to write proper schema for the following examples.

### Example 1
Given the following JSON response:
```json
[ "This", "is", "my", "array"]
```

The following properties would be made available in AppMap data:
```json
"return_value": {
  "class": "Array",
  "items": [ "class": "String" ]
 }
```

And can be translated to an OpenAPI compliant definition:
```yml
/api/example:
  get:
    responses:
      '200':
        content:
          application/json:
            schema:
              type: array
              items:
                type: string
```

### Example 2
Given the following JSON response:
```
{
  "keys": ["one", "two", "three"]
}
```

The following properties would be made available in AppMap data:
```json
"return_value": {
  "class": "Hash",
  "properties":  [ 
    {
      "name": "keys",
      "class": "Array",
      "items": [ { "class": "string" } ]
    }
  ]
}
```

And can be translated to an OpenAPI compliant definition:
```yml
/api/example:
  get:
    responses:
      '200':
        content:
          application/json:
            schema:
              type: object
              properties:
                keys: 
                  type: array
                  items:
                    type: string
```

### DIFF
--- a/lib/appmap/value_inspector.rb
+++ b/lib/appmap/value_inspector.rb
@@ -3,6 +3,7 @@ module AppMap
     extend self
 
     MAX_DEPTH = 3
+    MAX_ARRAY_ELEMENTS = 5
 
     def detect_size(value)
       # Don't risk calling #size on things like data-access objects, which can and will issue queries for this information.
@@ -11,10 +12,17 @@ module AppMap
       end
     end
 
-    def detect_schema(value, max_depth: MAX_DEPTH, type_info: {}, observed_values: Set.new(), depth: 0)
-      return type_info if depth == max_depth
+    def detect_schema(
+      value,
+      max_depth: MAX_DEPTH,
+      max_array_elements: MAX_ARRAY_ELEMENTS,
+      type_info: { class: best_class_name(value) },
+      observed_values: Set.new,
+      depth: 0
+    )
+      return type_info if depth >= max_depth && !array_like?(value)
 
-      if value.respond_to?(:keys)
+      if hash_like?(value)
         return if observed_values.include?(value.object_id)
 
         observed_values << value.object_id
@@ -23,17 +31,22 @@ module AppMap
           next_value = value[key]
 
           value_schema = begin
-              { name: key, class: best_class_name(next_value) }
-            rescue
-              warn "Error in add_schema(#{next_value.class})", $!
-              raise
-            end
+            { name: key, class: best_class_name(next_value) }
+          rescue
+            warn "Error in add_schema(#{next_value.class})", $!
+            raise
+          end
 
           detect_schema(next_value, **{ max_depth: max_depth, type_info: value_schema, observed_values: observed_values, depth: depth + 1 })
         end.compact
         type_info[:properties] = properties unless properties.empty?
-      elsif value.respond_to?(:first)
-        detect_schema(value.first, **{ max_depth: max_depth, type_info: type_info, observed_values: observed_values, depth: depth + 1 })
+      elsif array_like?(value)
+        type_info[:items] = value.take(max_array_elements).map do |next_value|
+          value_schema = { class: best_class_name(next_value) }
+          detect_schema(next_value, **{ max_depth: max_depth, type_info: value_schema, observed_values: observed_values, depth: depth + 1 })
+        end
+
+        type_info[:items] = type_info[:items].compact.uniq(&:hash)
       end
       type_info
     end
@@ -45,6 +58,16 @@ module AppMap
         value_cls = value_cls.superclass
       end
       value_cls&.name || "unknown"
+    end
+
+    private
+
+    def array_like?(value)
+      value.is_a?(Enumerable) && !hash_like?(value)
+    end
+
+    def hash_like?(value)
+      value.respond_to?(:keys)
     end
   end
 end
--- a/spec/schema_spec.rb
+++ b/spec/schema_spec.rb
@@ -60,12 +60,24 @@ describe 'Schema example' do
   describe 'Array of Hashes' do
     let(:value) { [ { id: 1, contents: 'some text' }, { id: 2 } ] }
     it 'is an array containing the schema' do
-      expect(schema).to match(hash_including(
-          properties: [ 
-            { name: :id, class: 'Integer' },
-            { name: :contents, class: 'String' }
+      expect(schema).to match(
+        hash_including(
+          class: 'Array',
+          items: [
+            {
+              class: 'Hash',
+              properties: [
+                { name: :id, class: 'Integer' },
+                { name: :contents, class: 'String' }
+              ]
+            },
+            {
+              class: 'Hash',
+              properties: [ { name: :id, class: 'Integer' } ]
+            }
           ]
-        ))
+        )
+      )
     end
   end
 end
--- a/spec/value_inspector_spec.rb
+++ b/spec/value_inspector_spec.rb
@@ -7,9 +7,11 @@ describe AppMap::ValueInspector do
   let(:parent_id) { nil }
   let(:exception) { nil }
   let(:max_depth) { nil }
+  let(:max_array_elements) { nil }
   let(:schema) {
     options = {}
     options[:max_depth] = max_depth if max_depth
+    options[:max_array_elements] = max_array_elements if max_array_elements
     AppMap::ValueInspector.detect_schema value, **options
   }
 
@@ -17,6 +19,7 @@ describe AppMap::ValueInspector do
     let(:value) { { id: 1, contents: "some text" } }
     it "is a one level schema" do
       expect(schema).to eq(
+        class: "Hash",
         properties: [
           { name: :id, class: "Integer" },
           { name: :contents, class: "String" },
@@ -29,6 +32,7 @@ describe AppMap::ValueInspector do
     let(:value) { { page: { page_number: 1, page_size: 20, total: 2383 } } }
     it "is a two level schema" do
       expect(schema).to eq(
+        class: "Hash",
         properties: [
           {
             name: :page,
@@ -46,6 +50,7 @@ describe AppMap::ValueInspector do
       let(:max_depth) { 1 }
       it "respects max depth" do
         expect(schema).to eq(
+          class: "Hash",
           properties: [
             {
               name: :page,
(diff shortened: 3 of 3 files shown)


spec/value_inspector_spec.rb
```diff
@@ -94,25 +165,31 @@
...
+                  { :class => "TrueClass", :name => "is_public" },
+                  { :class => "String", :name => "offering_key" },
+                  { :class => "NilClass", :name => "plan_type" },
+                  { :class => "Integer", :name => "price" },
+                  { :class => "NilClass", :name => "stripe_plan_id" },
+                  { :class => "NilClass", :name => "stripe_price_id" },
+                  { :class => "String", :name => "updated_at" }
+                ]
+              },
             ],
```
dustinbyrne, line 193: As far as backwards compatibility goes, I believe this is the only concerning change. Objects within an array now describe their type and any properties within a new object.

symwell: Tested this with Forem.  `openapi.yml` represented objects within an array with their type and any properties within a new object.

```
@@ -47632,8 +47905,18 @@ paths:
                       body_markdown:
                         type: array
+                        items:
+                          type: object
+                          properties:
+                            undefined:
+                              type: string
                       listing_category:
                         type: array
+                        items:
+                          type: object
+                          properties:
+                            undefined:
+                              type: string
```

```
@@ -47755,6 +48036,11 @@ paths:
                     properties:
                       listing_category:
                         type: array
+                        items:
+                          type: object
+                          properties:
+                            undefined:
+                              type: string
                   status:
```

kgilpin:
> Tested this with Forem and it preserved backward compatibility.
> 
> Strings were represented as objects.
> 
> ```
> @@ -47436,7 +47706,10 @@ paths:
>                  users:
>                    type: array
>                    items:
> -                    type: string
> +                    type: object
> +                    properties:
> +                      undefined:
> +                        type: string
> ```

This output of:

```
+                          properties:
+                            undefined:
+                              type: string
```

Looks like a show stopper to me (`undefined).

Also, we don't want `string` to become an object with an undefined property of type `string`.

kgilpin (changes requested):
Don't emit `undefined`. Don't covert `string` to an object with an `undefined` string property.

lib/appmap/value_inspector.rb
```diff
@@ -32,8 +34,12 @@ def detect_schema(value, max_depth: MAX_DEPTH, type_info: {}, observed_values: S
           detect_schema(next_value, **{ max_depth: max_depth, type_info: value_schema, observed_values: observed_values, depth: depth + 1 })
         end.compact
         type_info[:properties] = properties unless properties.empty?
-      elsif value.respond_to?(:first)
-        detect_schema(value.first, **{ max_depth: max_depth, type_info: type_info, observed_values: observed_values, depth: depth + 1 })
+      elsif value.respond_to?(:map)
+        type_info[:properties] = value.map do |next_value|
```
kgilpin, line 38: This could generate a large amount of data

spec/schema_spec.rb
```diff
@@ -60,12 +60,20 @@
...
     let(:value) { [ { id: 1, contents: 'some text' }, { id: 2 } ] }
     it 'is an array containing the schema' do
-      expect(schema).to match(hash_including(
-          properties: [ 
-            { name: :id, class: 'Integer' },
-            { name: :contents, class: 'String' }
+      expect(schema).to match(
+        hash_including(
+          class: 'Array',
+          properties: [
```
kgilpin, line 66: What does `class: Array` add, given that `properties` is already known to be an Array?

spec/value_inspector_spec.rb
kgilpin: The `undefined` here is a problem.

lib/appmap/value_inspector.rb
```diff
@@ -32,8 +34,12 @@ def detect_schema(value, max_depth: MAX_DEPTH, type_info: {}, observed_values: S
...
+          value_schema = { class: best_class_name(next_value) }
+          detect_schema(next_value, **{ max_depth: max_depth, type_info: value_schema, observed_values: observed_values, depth: depth + 1 })
+        end
+        type_info[:properties] = type_info[:properties].compact.uniq { |schema| schema[:class] }
```
kgilpin, line 42: I am not sure that this behavior of emitting an array of `properties` object for every array entry is really compliant with the appmap specification (at least, with the intent of it) or with the behavior of the OpenAPI generation tool. It might be better to do a merge of the `schema` here and avoid setting properties to an array. It's probably causing problems downstream because the OpenAPI generation code is not anticipating that this will happen?

lib/appmap/value_inspector.rb
dustinbyrne: I've added a limit (5)

lib/appmap/value_inspector.rb
dustinbyrne: I've switched to something a little more in line with OpenAPI - adding an `items` property which describes the type(s) held by the array.

spec/schema_spec.rb
dustinbyrne: Let me double check that I'm not emitting redundant type information before I respond to this.

dustinbyrne: Here's an example of the current state:
```json
{
  "return_value": {
    "class": "Array",
    "value": "[{mapset_id=>13, scenario_uuid=>03aee8be-b5a0-494a-8a87-a1732bc7dd29...",
    "object_id": 95540,
    "size": 20,
    "properties": [
      { "name": "mapset_id", "class": "Integer" },
      { "name": "scenario_uuid", "class": "String" },
      { "name": "code_objects", "class": "Array" },
      {
        "name": "metadata",
        "class": "Hash",
        "properties": [
          { "name": "app", "class": "String" },
          { "name": "git", "class": "Hash" },
          { "name": "name", "class": "String" },
          { "name": "client", "class": "Hash" },
          { "name": "language", "class": "Hash" },
          { "name": "recorder", "class": "Hash" },
          { "name": "timestamp", "class": "Float" },
          { "name": "frameworks", "class": "Array" }
        ]
      },
      { "name": "labels", "class": "Array" }
    ]
  }
}
```

The `return_value` is an array and `properties` describes the schema of the object contained within. The above schema implicitly implies everything within the array is an object. We'll miss primitives, mixed types or nested arrays.

Here's an example of an API which returns the JSON response `[1, 2, 3]`:
```json
{
  "return_value": { "class": "Array", "value": "[1, 2, 3]", "object_id": 99920, "size": 3 },
}
```

We lose the type information because we're not capturing it. This is why `class: Array` is specified at this level. To compare, here are the previous outputs after the changes in this PR:

```json
{
  "return_value": {
    "class": "Array",
    "value": "[{mapset_id=>13, scenario_uuid=>03aee8be-b5a0-494a-8a87-a1732bc7dd29...",
    "object_id": 378440,
    "size": 20,
    "items": [
      {
        "class": "Hash",
        "properties": [
          { "name": "mapset_id", "class": "Integer" },
          { "name": "scenario_uuid", "class": "String" },
          { "name": "code_objects", "class": "Array", "items": [] },
          {
            "name": "metadata",
            "class": "Hash",
            "properties": [
              { "name": "app", "class": "String" },
              { "name": "git", "class": "Hash" },
              { "name": "name", "class": "String" },
              { "name": "client", "class": "Hash" },
              { "name": "language", "class": "Hash" },
              { "name": "recorder", "class": "Hash" },
              { "name": "timestamp", "class": "Float" },
              { "name": "frameworks", "class": "Array", "items": [{ "class": "Hash" }] }
            ]
          },
          { "name": "labels", "class": "Array", "items": [] }
        ]
      }
    ]
  }
}
```

```json
{
  "return_value": {
    "class": "Array",
    "value": "[1, 2, 3]",
    "object_id": 94980,
    "size": 3,
    "items": [{ "class": "Integer" }]
  }
}
```
It's more verbose but it's necessary for us to describe the element type of any array. It could be argued that any property object containing `properties` or `items` doesn't need `class` specified (in this case we know it's an `object` or `array`) but I'd err on the side of retaining more data.

dustinbyrne:
Changing Array descriptors from being stored in `properties` to `items` should mean this is fully backward compatible.

symwell:
Some observations in appmap-server that may help.

1) There's a `ref_0` in `labels`.  Is this ok?  I [saw](https://github.com/getappmap/appmap-js/pull/995#issue-1570604769) there can be a `ref_0` in `code_objects`.
```
+                        labels: &ref_0
+                          type: array
+                          items:
+                            anyOf:
+                              - type: array
+                              - type: object
+                              - type: integer
+                              - type: number
+                              - type: boolean
+                              - type: string
```
Also saw this. Does `*ref` mean something different than `&ref`?
```
-                appmaps:
-                  type: array
+                appmaps: *ref_0
```

2) In diffing old openapi.yml with new, `multipart/mixed` is missing from new.
```
-          multipart/mixed:
-            schema:
-              type: object
-              properties:
```

3) `/` isn't in `openapi.yml` anymore.
```
-  /:
-    get:
-      responses:
-        '200':
-          content:
-            text/html: {}
-          description: OK
-        '303':
-          content: {}
-          description: See Other
```

4) Perhaps (3) is a more general issue, many API calls are missing.  In Forem no API calls are missing.
```
-  /api_keys:
-    post:
-      responses:
-        '302':
-          content: {}
-          description: Found

-  /api_keys/new:
-    get:
-      responses:
-        '200':
-          content:
-            text/html: {}
-          description: OK

-  /api_keys/{id}:
-    delete:
-      responses:
-        '302':
-          content: {}
-          description: Found

-  /application_factory:
-  /applications/new:
-  /configuration:
-  /license:
-  /lobby:
-  /scenario_uploads/new:
-  /scenarios:
...
```

In forem:

5) `locale` is now generated.
```
+        - name: locale
+          in: query
+          schema: {}
```

6) There's still `undefined`. I didn't see any `undefined` in appmap-server.
```
+                type: object
+                properties:
+                  undefined:
+                    type: object
```
7) In ` (/locale/{locale})/api/listings:` some properties are missing.  In other API calls I didn't see properties missing.
```
-                  properties:
-                    body_markdown:
-                      type: string
-                    category:
-                      type: string
-                    organization_id:
-                      type: integer
-                    title:
-                      type: string
```

dustinbyrne:
These issues should be addressed in https://github.com/getappmap/appmap-js/pull/995. Regarding missing routes, I'm not sure. The changes here are purely additive at this point.

lib/appmap/value_inspector.rb
```diff
@@ -46,5 +64,16 @@ def best_class_name(value)
       end
       value_cls&.name || "unknown"
     end
+
+    private
+
+    def array_like?(value)
+      value.is_a?(Enumerable) && !value.is_a?(Hash)
```
kgilpin, line 71: Would you use `!hash_like?(value)`?

lib/appmap/value_inspector.rb
```diff
@@ -23,17 +31,27 @@ def detect_schema(value, max_depth: MAX_DEPTH, type_info: {}, observed_values: S
...
           detect_schema(next_value, **{ max_depth: max_depth, type_info: value_schema, observed_values: observed_values, depth: depth + 1 })
         end.compact
         type_info[:properties] = properties unless properties.empty?
-      elsif value.respond_to?(:first)
-        detect_schema(value.first, **{ max_depth: max_depth, type_info: type_info, observed_values: observed_values, depth: depth + 1 })
+      elsif array_like?(value)
+        # We can't just cut off the recursion here, because we need to know the
+        # class of the elements in the array. Allow one more level of recursion
+        # to get the class of the elements.
+        # next_max_depth = depth + 1 == max_depth ? max_depth + 1 : max_depth
```
kgilpin, line 47: Did you mean to leave this in?

lib/appmap/value_inspector.rb
kgilpin: That way, anything Enumerable is either array_like or hash_like.

lib/appmap/value_inspector.rb
```diff
@@ -11,10 +12,17 @@ def detect_size(value)
...
-      return type_info if depth == max_depth
+    def detect_schema(
+      value,
+      max_depth: MAX_DEPTH,
+      max_array_elements: MAX_ARRAY_ELEMENTS,
+      type_info: { class: best_class_name(value) },
+      observed_values: Set.new,
+      depth: 0
+    )
+      return type_info if depth >= max_depth && !array_like?(value)
```
kgilpin, line 23: Not that it's a big issue, but it will dive deeply if there are Arrays of Arrays of Arrays of Arrays...

spec/value_inspector_spec.rb
```diff
@@ -61,9 +66,154 @@
...
+
+      it "won't exceed max_depth when inspecting a Hash" do
+        expect(schema).to eq(
+          class: "Array",
+          items: [
+            {
+              class: "Hash",
+              properties: [
+                { name: :id, class: "Integer" },
+                { name: :children, class: "Array", items: [ { class: "Hash" } ] }
```
kgilpin, line 144: Is it valid to not have any schema on this hash?

spec/value_inspector_spec.rb
```diff
@@ -61,9 +66,154 @@
...
+        ],
+      )
+    end
+  end
+
+  describe "Max array elements" do
+    let(:value) { [ 1, 'two', { id: 3 }, [ 'four' ] ] }
+    let(:max_array_elements) { 3 }
+
+    it "only describes the first N elements" do
```
kgilpin, line 183: I think that the schema of the fourth element would be merged in with the schema of the second, due to the `uniq`, so I don't think this test actually verifies the truncation at N=4.

lib/appmap/value_inspector.rb
kgilpin: I see there's a test for this later 👍

lib/appmap/value_inspector.rb
dustinbyrne: Nope, thanks!

spec/value_inspector_spec.rb
dustinbyrne: This is essentially the same as the following, which would be valid:
```yml
type: array
items: object
```

lib/appmap/value_inspector.rb
dustinbyrne: It's possible, but it'll stop at the first `object` beyond or at `max_depth`.

So with a `max_depth` of `1`, this'll stop at the first object:
```
[ { children: [ { children: [ { children: [ { children: [ { children: [ { children: [ { children: []]]]]]]
```

The only real way to abuse this would be if you for some reason had deeply nested primitives, e.g.:
```
[[[[[[[[[[1]]]]]]]]]]
```

spec/value_inspector_spec.rb
dustinbyrne: This'll end up resolving to
```yml
class: Array
items:
  - class: Integer
  - class: String
  - class: Hash
    properties:
      - name: id
        class: Integer
  - class: Array
    items:
      - class: String
```

The `uniq` is using each items `hash` which should be unique in this case.

kgilpin (approved):

PR #320, opened 2023-03-09
fix: Report minitest test status

kgilpin:
AppMap `test_status` is not reporting `failed` when a Minitest fails. This is the fix for that.


apotterri (approved):
LGTM

PR #326, opened 2023-05-02
Hook Rack in Rails

### LINKED ISSUE #323: Error occurs when passing bad credentials to devise.
Hello!

This is occurring in a rails app running rails 7.0.1 and devise 4.8.1. When entering bad credentials (blank, for example), the expected result is that it should display the devise error message 'Failed to login'. The actual result is that an error occurs at this line in appmap: 

https://github.com/getappmap/appmap-ruby/blob/master/lib/appmap/handler/rails/request_handler.rb#L76

The error that I see:
```
NoMethodError in Admins::SessionsController#create
undefined method `status' for {:controller=>"Admins::SessionsController", :action=>"create", :request=>#<ActionDispatch::Request POST "http://localhost:3000/admins/sign_in" for 127.0.0.1>, :params=>{"authenticity_token"=>"[FILTERED]", "admin"=>{"username"=>"", "password"=>"[FILTERED]", "remember_me"=>"0"}, "commit"=>"Sign In", "controller"=>"admins/sessions", "action"=>"create"}, :headers=>#<ActionDispatch::Http::Headers:0x000000010e1ead98 @req=#<ActionDispatch::Request POST "http://localhost:3000/admins/sign_in" for 127.0.0.1>>, :format=>:html, :method=>"POST", :path=>"/admins/sign_in", :view_runtime=>nil, :db_runtime=>4.178000211715698}:Hash
```
To recreate you'll need to have an app that uses devise for authentication and add the appmap gem to your Gemfile. Install and restart the server. Then, after signing out, enter bad credentials.

dividedmind:
Hook requests in Rails on the Rack level in addition to ActionController. This allows intercepting any changes and responses made by the middleware. We're still hooking ActionController, because controller tests don't use Rack, but the ActionController hook is skipped if the rack hook has been activated on the request.

Fixes #323 (I believe)

### DIFF
--- a/lib/appmap/handler/rails/context.rb
+++ b/lib/appmap/handler/rails/context.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'appmap/handler'
+
+module AppMap
+  module Handler
+    module Rails
+      # Context of a rails request tracking.
+      # Mostly a utility class to clean up and deduplicate request handler code.
+      class Context
+        def initialize(environment = nil)
+          environment[REQUEST_CONTEXT] = self if environment
+          @thread = Thread.current
+        end
+
+        def self.from(environment)
+          environment[REQUEST_CONTEXT]
+        end
+
+        def self.create(environment)
+          return if from environment
+
+          new environment
+        end
+
+        def self.remove(env)
+          env[REQUEST_CONTEXT] = nil
+        end
+
+        def find_template_render_value
+          @thread[TEMPLATE_RENDER_VALUE].tap do
+            @thread[TEMPLATE_RENDER_VALUE] = nil
+          end
+        end
+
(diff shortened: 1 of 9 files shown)


dividedmind:
Note something breaks the build, I'm trying to figure it out, I've switched it to draft.

apotterri:
> > Code looks good to me.
> > It seems worth installing it in an app that uses Devise, though, and verifying that the user's problem is actually fixed.
> > This seems like a possibility: https://github.com/ManickYoj/rails7-devise-template, or maybe one of these would do: https://henrytabima.github.io/rails-setup/docs/devise/example-applications .
> 
> I did (I created my own test app from scratch) and it does seem to solve the issue. I was reluctant to add it to tests so as not to complicate them unnecessarily.

Ok, that's fine. Your initial comment didn't seem all that confident, so I just wanted to make sure.

Thanks!

PR #328, opened 2023-05-08
fix: Handle KeyError in signature lookup

kgilpin:
Encountered while mapping forem:

```
An error occurred while loading ./spec/models/reaction_category_spec.rb.
Failure/Error: end

TypeError:
  no implicit conversion of Hash into Integer
WARNING: Shared example group 'Taggable' has been previously defined at:
  /home/runner/work/forem/forem/spec/models/shared_examples/taggable_spec.rb:1
...and you are now defining it at:
  /home/runner/work/forem/forem/spec/models/shared_examples/taggable_spec.rb:1
The new definition will overwrite the original one.
```

### DIFF
--- a/spec/method_hash_key_spec.rb
+++ b/spec/method_hash_key_spec.rb
@@ -1,0 +1,26 @@
+require 'spec_helper'
+require 'appmap/hook/method'
+
+describe 'method_hash_key' do
+  describe 'of a normal method' do
+    it 'is a hash' do
+      expect(AppMap::Hook.method_hash_key(AppMap::Hook, AppMap::Hook.method(:method_hash_key))).to be_a Integer
+    end
+  end
+  describe 'when the class hash raises an error' do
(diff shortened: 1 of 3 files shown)


spec/method_hash_key_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+require 'appmap/hook/method'
+
+describe 'method_hash_key' do
+  describe 'of a normal method' do
+    it 'returns a hash' do
+      expect(AppMap::Hook.method_hash_key(AppMap::Hook, AppMap::Hook.method(:method_hash_key))).to be_a Integer
+    end
+  end
+  describe 'when the class hash raises an error' do
+    it 'returns a hash' do
```
apotterri, line 11: It doesn't return a hash, though, does it?

apotterri (approved):
One nit, otherwise looks good.

PR #333, opened 2023-07-11
feat: Add `appmap-agent-config` command to bootstrap appmap.yml

dustinbyrne:
This command will be used during the AppMap manual install process to initialize an `appmap.yml`.

### DIFF
--- a/exe/appmap-agent-config
+++ b/exe/appmap-agent-config
@@ -1,0 +1,31 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'optparse'
+require 'appmap'
+require 'appmap/command/agent_setup/config'
+
+@options = { config_file: AppMap::DEFAULT_CONFIG_FILE_PATH, force: false }
+
+OptionParser.new do |parser|
+  parser.banner = 'Usage: appmap-agent-config [options]'
+
+  description = "AppMap configuration file path (default: #{AppMap::DEFAULT_CONFIG_FILE_PATH})"
+  parser.on('-c', '--config=FILEPATH', description) do |filepath|
+    @options[:config_file] = filepath
+  end
+
+  parser.on('-f', '--force', 'Overwrite existing configuration file') do
+    @options[:force] = true
+  end
+end.parse!
+
+begin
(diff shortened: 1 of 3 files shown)


exe/appmap-agent-config
```diff
@@ -0,0 +1,19 @@
...
+# frozen_string_literal: true
+
+require 'optparse'
+require 'appmap'
+require 'appmap/command/agent_setup/config'
+
+@options = { config_file: AppMap::DEFAULT_CONFIG_FILE_PATH }
+
+OptionParser.new do |parser|
+  parser.banner = 'Usage: appmap-agent-init [options]'
```
kgilpin, line 11: Command name needs to be updated here

lib/appmap/command/agent_setup/config.rb
```diff
@@ -0,0 +1,30 @@
...
+module AppMap
+  module Command
+    module AgentSetup
+      ConfigStruct = Struct.new(:config_file)
+
+      class Config < ConfigStruct
+        def perform
+          if File.exist?(config_file)
+            puts "AppMap configuration file already exists at #{config_file}"
+            exit(1)
```
kgilpin, line 15: I feel like process exit should be handled in the `exe` package? So maybe just raise or return false.

exe/appmap-agent-config
```diff
@@ -0,0 +1,19 @@
...
+
+  description = "AppMap configuration file path (default: #{AppMap::DEFAULT_CONFIG_FILE_PATH})"
```
kgilpin, line 13: Probably want to update this to indicate that the file will be created by this command, that it won't be overwritten though. Maybe a `-f, --force` option to replace it? That would enable a developer to use it to "reset" the configuration.

dustinbyrne: Good idea. I've added the `--force`/`-f` option to overwrite any existing config.

lib/appmap/command/agent_setup/config.rb
dustinbyrne: I've changed it to raise an error

kgilpin (approved):

PR #351, opened 2024-01-31
Minitest tagz 20240130

### LINKED ISSUE #350: Minitest recording can be selectively disabled using tags
Similar to [the RSpec feature](https://appmap.io/docs/reference/appmap-ruby.html#enable-and-disabling-appmaps), enable minitest test cases to be selectively disabled via integration with https://github.com/jbodah/minitest-tagz.

apotterri:
Check to see if Minitest::Tagz has been used the add the tag :noappmap
to a test method. If so, don't generate an AppMap for the test.

Fixes #350 .

### DIFF
--- a/lib/appmap/minitest.rb
+++ b/lib/appmap/minitest.rb
@@ -85,14 +85,24 @@ module AppMap
 
         @recording_count += 1
 
-        @recordings_by_test[test.object_id] = Recording.new(test, name)
+        recording = if defined?(::Minitest::Tagz) && disabled_by_tag(test, name)
+            :disabled
+          else
+            Recording.new(test, name)
+          end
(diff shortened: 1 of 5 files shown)


apotterri:
Kind of unfortunate that `semantic-release` requires the ominous-sounding "BREAKING CHANGE" in the comment to bump to v1.0.0....

kgilpin:
Yes that’s why we are still on v0 :-)

apotterri:
I still think we should do it (and probably should have done it a while ago). It's quite stable, and just continuing to increment the minor version seems weird....

kgilpin:
Yeah me too.

kgilpin (approved):

PR #359, opened 2024-04-24
fix: Devise incompatibility

kgilpin:
See https://github.com/getappmap/appmap-ruby/issues/358


lib/appmap/handler/rails/test_route.rb
```diff
@@ -0,0 +1,26 @@
...
+        protected
+
+        @test_route_warned = false
+
+        def test_route_warn(request)
+          return if @test_route_warned
+
+          @test_route_warned = true
+          warn "Notice: Failed to match route for #{request&.path_info || "an unknown path"}: #{$ERROR_INFO}"
+          warn "Notice: A solution for this problem is forthcoming, see https://github.com/getappmap/appmap-ruby/issues/360"
```
dividedmind, line 21: I'd also mention the consequence: that the normalized route will not be resolved.

dividedmind (approved):
Looks good! Thank you for splitting the formatting changes to a separate commit.

PR #361, opened 2024-05-20
feat: Ruby 3.3.1

kgilpin:
Try adding 3.3.1 to the test matrix

### DIFF
--- a/ext/appmap/appmap.c
+++ b/ext/appmap/appmap.c
@@ -1,6 +1,7 @@
 #include <ruby.h>
 #include <ruby/debug.h>
 #include <ruby/intern.h>
+#include <ruby/version.h>
 
 // Seems like CLASS_OR_MODULE_P should really be in a header file in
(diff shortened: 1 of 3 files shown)


kgilpin:
Possibly helpful - https://rubyreferences.github.io/rubychanges/3.3.html

ext/appmap/appmap.c
```diff
@@ -17,7 +18,13 @@ static VALUE
 singleton_method_owner_name(VALUE klass, VALUE method)
 {
   VALUE owner = rb_funcall(method, rb_intern("owner"), 0);
+
+#if RUBY_API_VERSION_CODE < 30200
```
kgilpin, line 22: Whoa

apotterri: It was pretty grotty before. The change they made in 3.2 allows it to be slightly less so.

apotterri (approved):

